PR #3, opened 2019-06-18
Fix tests and setup for CPU development

JelleAalbers:
This:
  * "Ports the flamedisx tests to tensorflow"... which in practice means adding '.numpy()' after the flamedisx calls, so we can then use numpy for the tests.
  * Removes the depedency on straxen by duplicating the necessary pieces of code. When straxen moves away from tensorflow 1, we can change this (but that could be a while).
  * Fix a few bugs that made the tests fail, but did not cause a visible problem in the GPU tests in colab:
    * lookup_axis1 used out-of-bounds indices
    * Added a few numpy_out and .values to prevent pandas Series from messing up simulate_nq. On master this code currently causes a core dump.
    * The bound of {quantum}_produced was calculated so weirdly that it could come out lower than the MLE. This was likely present already before the tensorflow port, but it manifested only now (perhaps some rounding has changed). The calculation is still weird, but at least now it passes this minimal test.
  * Remove a few unused lines (likely present before the port)
  * Add a very crude batch system back in: just a for loop over set_data and likelihood. Without this I can't compute the likelihood for a reasonable number of events on my poor 8GB RAM laptop.
  * Add TravisCI

@pelssers / @CristianAntochi could i ask one of you to check that flamedisx still works fine on colab? The setup might be easier now that the requirements.txt is updated, although you'd still have to install tensorflow-gpu manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: python
+dist: xenial
+python:
+- '3.6'
+- '3.7'
+install:
+- pip install -r requirements.txt
+script:
+- python setup.py test
+deploy:
+  provider: pypi
+  skip_existing: true
+  user: jelleaalbers
+  password:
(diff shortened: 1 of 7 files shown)


pelssers:
Great!

- I've updated the colab notebook as well and added it to the PR, basically removing straxen and checking the TF2 beta1 release.
- The likelihood computation runs slightly slower than before, maybe due to the _lookup_axis1 changes.
- Maybe we can add tests where we compare the types and shapes of  `gimme(numpy_out=True)` to `gimme().numpy()`.

JelleAalbers:
Thanks for checking; the slowdown could also come from the change in the bounds computation.

Issue #4, opened 2019-06-18
Add tests of gimme with and without numpy_out


JelleAalbers:
@pelssers mentioned in #3: add tests where we compare the types and shapes of gimme(numpy_out=True) to gimme().numpy()

JelleAalbers:
Added in #5

PR #5, opened 2019-06-18
Port remaining functions to tf, rearrange things

JelleAalbers:
This adds the following:
  * Port the remaining functions in ``x1t_sr0.py``, and the core function ``_dimsize``, to tensorflow. All functions called in ``likelihood`` should now be pure tensorflow.
    * The S1 and S2 relative light yields are now evaluated once during set_data, rather than every time during a likelihood call. This is done by the new function ``add_extra_columns``.
  * ``likelihood`` is now a ``@tf.function`` function that returns tensors. It does not use batching and only takes the likelihood parameters as arguments. It seems to work as expected (first call is slow, next ones faster), though I've only been able to test this with a few events on my laptop.
  * Move the crude batching to a new function ``batched_likelihood``, which returns numpy arrays. This evaluates likelihoods for all events by splitting the data into batches and repeatedly calling ``set_data``. It does not use the `@tf.function`'ed likelihood, since it seems this does not refresh its inputs as expected during the batching.
  * Move the annotation of the data (bound estimation etc.) to ``annotate_data``. Conversely, ``set_data`` calls ``annotate_data``, unless the option ``annotated=True`` is passed.
  * Move some utility functions into ``utils.py``, and use the ``@export`` mechanism throughout flamedisx.
  * Clean up gimme and add tests described in (#4)

@pelssers / @CristianAntochi  I would again be grateful if you can see how this affects things in colab.

### DIFF
--- a/flamedisx/__init__.py
+++ b/flamedisx/__init__.py
@@ -1,4 +1,6 @@
 __version__ = '0.0.1'
 
-from . import core
-from .core import ERSource, NRSource
+from .utils import *
+from .xenon_corrections import *
+from .core import *
+from .x1t_sr0 import *
--- a/flamedisx/core.py
+++ b/flamedisx/core.py
@@ -1,13 +1,17 @@
 import inspect
 
+from multihist import Hist1d
+import numpy as np
+import pandas as pd
+from scipy import stats
 import tensorflow as tf
 import tensorflow_probability as tfp
 # Remove once tf.repeat is available in the tf api
 from tensorflow.python.ops.ragged.ragged_util import repeat
-import numpy as np
-from scipy import stats
-import pandas as pd
-from multihist import Hist1d
+from tqdm import tqdm
+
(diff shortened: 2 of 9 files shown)


pelssers:
When running the notebook on colab it is no longer possible to use all events at once, we now run into OOM errors.
I guess the best thing for the long term is adding a batch dimension to the data.
Another thing we can check is if we can go back to float16 for almost all computation, the only place where we know we really need float32 is for the beta-binomial.

I've checked that batched_likelihood still works fine on colab.

Using `@tf.function` doesn't work out-of-the-box on our complex likelihood function. We have to make sure TF detects when we change the state (set new data). I've been reading [this](https://github.com/tensorflow/community/blob/master/rfcs/20180918-functions-not-sessions-20.md)

JelleAalbers:
Thanks for checking! Yes, to deal with larger datasets we will have to add a proper batching system.

PR #7, opened 2019-07-09
Dev ll comp

CristianAntochi:
I have modified the code to allow batching.
It both batches the data and the tensor cache with one call of set_data.
The bestfit function loops over the batches and computes the likelihood and the gradient and the inverse_hessian computes the hessian over the batches and sums it up, then the inverse hessian is computed.


JelleAalbers:
Thanks a lot Christian! This is the missing piece of the puzzle that allows us to fit larger datasets. Nice work.

I added a 'setup.cfg' file, whose absence caused pytest to miss our tests. Not sure how they ran on my laptop before, or why they ran for some commits in this branch but not for others...

Anyway, I'll try to fix the tests and maybe clean up gimme a little bit, then we can merge this.

JelleAalbers:
Here's a few of the changes I made:
  * Remove `batched_data` in favor of putting every numerical column from data in to tensor_cache.
  * User specifies batch_size (in events), not number of batches, when initializing a LogLikelihood. This is tracked throughout.
  * Any function that takes i_batch can also take i_batch=None, producing results for the entire dataset.
  * Refactor gimme, introducing a new poorly named function '_fetch', which fetches from the tensor cache.
  * Fix an old bug: log10 -> log likelihood. This would have caused wrong bestfit rate multipliers and incorrect errors/correlations. I didn't check whether the ones we have now /are/ correct though :-)

I also took the opportunity to do refactor flamedisx a bit more generally:
  * No more `set_data`! You now pass the data, along with the initial parameters, into the `__init__`. This means the source instances don't have any state they (and you) need to track anymore. (There are some "_skip..." options to `__init__` which create partially functioning instances, which is not so nice, but these are only used internally).
  * Methods that took separate data/params arguments, such as `simulate`, `estimate_mu` and `mu_interpolator`, are now class methods: you can call them directly on ERSource/NRSource. Internally they will make an instance with the right data/parameters (multiple times, actually, in simulate's case).
  * source.py now contains only a bare-bones Source class. The actual model function defaults are factored out to a new file er_nr_base.py.

Unfortunately, I am now no longer able to use @tf.function on the objective (or anywhere, really): it complains about 'None' dimensions showing up, originating all the way down to dimsize. This was an important optimization for the GPU, so I hope we can figure out what I did wrong and fix it -- it probably worked in @CristianAntochi's original version, though I wasn't able to run that directly on the CPU.

As for speed, on my laptop (using 3-4 CPUs), with 18 events (1/32 of the SR) radon data:
  * bestfit takes ~40 sec
  * inverse_hessian takes ~137 sec

So a full CPU SR0 ER fit would likely take 32x this time, i.e. 21 minutes for bestfit, then 73 minutes for the hessian: 1.5 hours in total. And this is only at max_sigma=3, at =5 it would likely take 2-3x longer. So we definitely need the GPU, or more optimization :-) Batch size seems to be irrelevant to the CPU performance.

JelleAalbers:
@CristianAntochi could you test if this still runs on colab, and if you are happy with these changes? Thanks again for the work in implementing this. Let's have a look next week on how we get it faster, e.g. with tf.function.

JelleAalbers:
I tried the `test_ll` notebook in a midway GPU job (Tesla K80, using just one GPU). Here are the results:

| Device         | N events | Batch size | Bestfit (s) | evt/s | Hessian (s) | evt/s |
|----------------|----------|------------|-------------|-------|-------------|-------|
| Jelle's laptop | 18       | 1          | 40          | .45   | 137         | .13   |
|                | 18       | 5          | 40          | .45   | 137         | .13   |
| K80 (midway)   | 40       | 1          | 101         | 0.4   | 137         | .29   |
|                | 160      | 10         | 138         | 1.15  | 88          | 1.8   |
|                | 160      | 40         | 80          | 2     | 41          | 4     |
|                | 160      | 80         | 72          | 2.2   | crash       | crash |
|                | 160      | 160        | 88          | 1.8   | crash       | crash |

There was some (10-20%) jitter in the times, probably because other people were using other GPUs on the same node at the same time.

It seems a batch size of ~40 is good. Perhaps the best is just to set it as high as possible before the GPU runs out of memory (crash means ResourceExhaustedError).

Overall the CPU -> GPU speedup is about 10x, if I count my laptop as 2 CPUs. There is probably a lot of room left for optimization: BBF reported a 100-1000x speedup (p.220 of Matt's thesis).

pelssers:
I've added a commit updating the flamedisx_tf2 notebook. It now runs again on colab (still without Autograph, I'll look into that next).

pelssers:
By precomputing the dimsizes when initializing the source autograph mode works again. I'm seeing the same 4x speedup when swithing on autograph when computing differential rate on colab.

JelleAalbers:
Thanks for the fix @pelssers! Here are the new benchmarks. I'm still using only one of the K80's GPUs.

| Device         | N events | Batch size | Bestfit (s) | evt/s | Hessian (s) | evt/s |
|----------------|----------|------------|-------------|-------|-------------|-------|
| Jelle's laptop |  18      |  1         | 66          | 0.27  | 137         | 0.13  |
|                |  18      | 18         | 44.2        | 0.41  | swapping    | ?     |
| K80 (midway)   | 160      | 10         | 71          | 2.2   | 81          | 2.0   |
|                | 160      | 40         | 40          | 4.0   | 41          | 4.0   |
|                | 160      | 60         | 38          | 4.2   | 49          | 3.2   |
|                | 160      | 80         | 33          | 4.8   |  crash     |  crash  |

This speeds up the bestfit by a factor 2 on the GPU; now we have a ~20x speedup compared to my one of my laptop's CPUs. One might think there is still some room left for optimization, but I wouldn't know an easy entry point.

Since tf.function has been re-enabled, I did these benchmarks after calling doing ll.bestfit(guess), to ensure all relevant graphs get traced. The Hessian still does not use tf.function (it never worked with that, not even before batching), so its performance is, as expected, unchanged.

Notice the performance on my laptop went down with batch_size=1, and gets back to its pre-tf.function level when using a large batch size. Not sure about this. Could it be tracing every batch? But (a) why would it need to, since the dimsizes are always the same, and (b) why would it still be slower after the first bestfit call?

Anyway, I think it's merge time for this PR.

JelleAalbers:
Oops, I hit merge instead of squash merge... fixing...

PR #9, opened 2019-07-18
Improving Hessian computation

pelssers:
I've rewritten the `inverse_hessian` method, it is now cleaner and faster.

Instead of calculating the matrix element by element it is now computed row by row using only Tensorflow functions and one list comprehension. All Numpy code is removed, batching is preserved.

Even though this implementation computes the entire Hessian matrix and does not exploit the fact the matrix is symmetric it is still about 5x to 8x faster than the previous implementation. Running the test_ll notebook on Colab I see the time for one function call go down from 55s to around 8s on a Colab GPU node).

I've tried running the function with Autograph but this results in an error. Although that may also be related to `minus_ll()` which does not run with Autograph either.

If `.hessians` is added to `GradientTape` is a future TF release we might switch to using that instead.

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -206,7 +206,7 @@ class LogLikelihood:
 
         hessian = tf.zeros((n, n), dtype=fd.float_type())
 
-        for i_batch in tf.range(self.n_batches):
+        for i_batch in range(self.n_batches):
             with tf.GradientTape(persistent=True) as t2:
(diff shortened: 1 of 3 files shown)


pelssers:
I've added a fix for an edge case when the number of events in the last batch is not a multiple of the batch size. Secondly we can now use -1 as the last batch number. Thanks @CristianAntochi for spotting it.

PR #13, opened 2019-08-13
Optimize function tracing

pelssers:
We use Tensorflow's `tf.function` to trace several performance critical functions (in particular `_differential_rate` in `Source`). These functions currently take Python arguments and depend on Python state to work properly. This means that these functions are retraced when the arguments change. This makes `batched_differential_rate` very slow and effects overall performance as well.
This PR rewrites the source and inference class to use a single trace of `_differential_rate` instead of triggering a retrace for each batch in each call to `batched_differential_rate`.

To make this work 2 things need to be changed:
1. `i_batch` needs to be a tensor instead of a Python integer.
   - This has the side effect that we can no longer look up the batch tensor in the `tensor_cache_list`. Instead the `tensor_cache_list` is replaced by the `data_tensor` which has shape `(n_batches, events_per_batch, n_column)`. The `_fetch` method still retrieves the right slice from the `data_tensor`. The conversion from column key to index is done via a `tf.lookup` lookup table.

2. `params` (or `ptensor`) needs to be part of the input signature as well.
   - Before we could simply set `self._params = params` before calling `_differential_rate`. But when using a single trace these values will not be updated. Instead we pass `ptensor` to `_differential_rate`. To determine the shape of `ptensor` the likelihood class sets `param_names` for each source. When `differential_rate` is called for the first time the function is traced with an input signature determined from `param_names` to ensure the right shape `ptensor` can be passed.

Running the flamedisx_tf2 notebook on a Colab GPU node we see `batched_diferential_rate` now works as expected (Tracing only for the first batch instead of each batch).

Running the Tutorial notebook on a Colab GPU node, we observe about a factor 2 speedup (without tracing time) for `bestfit` with a 2 parameter model. Hessian computation runs as fast as before.


pelssers:
Thanks, yes, we should look into optimizer stability next and see if we can improve gradient computation (which should reduce memory usage as well).

PR #19, opened 2019-09-11
Bfgs optimizer using hessian

pelssers:
This implements the bfgs optimizer instead of the lbfgs optimizer for bestfit. The reason for this change is that we can pass an estimate of the inverse hessian matrix to the optimizer to increase its performance.
Bestfit now has a boolean flag `use_hessian` which selects if the inverse hessian is used by the optimizer or not.
The hessian matrix is explicitly symmetrized such that it passes a check in `bfgs_minimize`. Before it failed since the matrix was not symmetric to the 7th decimal.

Performance of `_log_likelihood_grad2` could be improved by only computing the upper triangular part of the hessian since the calls to tf.gradient are expensive. On the other hand, the hessian is only evaluated once on the guess and once on the best fit for a single fit.

Tests have been split off into `test_source` and `test_inference`. Tests for the hessian and bestfit are added to `test_inference`.


JelleAalbers:
Nice work Bart! Since a main motivation of the new optimizer is performance, do you have an idea of how much faster e.g. our benchmarks from the tutorial become?

pelssers:
Running the test_ll notebook I see:
`bestfit(guess)` 3.95s on master
`inverse_hessian(bestfit)` 3.5s on master

`bestfit(guess)` 4.63s on bfgs branch (which includes the hessian calculation, implying that bestfit itself only takes about 1.1 seconds if hessian stayed the same.

I'll run the tutorial notebook next with a bigger fit for a more robust result.
(And fix the symmetry check that currently breaks Travis)

pelssers:
Here are the benchmarks from running the Tutorial notebook. In all cases 1000 events are simulated (840 and  858 events returned for the master and bfgs branch respectively), with batch_size 100.

Another difference between the two branches is that the proposed bfgs branch no longer traces the objective function passed to the optimizer. The reason for this is that at the time of  tracing the guess is frozen in and a new likelihood object needs to be created to use a new guess. This is not a problem when doing one fit (calling bestfit once) but this was not obvious and we definitely want to use multiple bestfit calls in the future. For small datasets/models this actually speeds up the bestfit since the overhead from the tracing is removed. For large datasets/models there does not seem to be a big difference.
'Trace' in the below tables includes tracing of the likelihood and objective in the master branch. In the bfgs branch only the likelihood trace is included.
'Fit' means one bestfit call, which also includes a call to inverse_hessian in the bfgs branch.
'Summary' also calls inverse_hessian (but at the bestfit position instead of the initial guess position).

2 param model | Master branch | BFGS branch
---|--------------------|------------------
Trace + Fit + Summary | 29.4s | 26.1s
Bestfit (no hessian) | **17.8s** | 16s
Bestfit (using hessian) | - | **17s**
Inverse_hessian | 3.96s | 3.7s
Bestfit (using hessian) - hessian | - | **13.3s**

6 param model | Master branch | BFGS branch
---|--------------------|------------------
Trace + Fit + Summary | 39.1s | 32s
Bestfit (no hessian) | **20.6s** | 20s
Bestfit (using hessian) | - | **17.3s**
Inverse_hessian | 11.5s | 10.8s
Bestfit (using hessian) - hessian | - | **6.5s**

Using the hessian in the bestfit significantly speeds up the bestfit, especially for the 6 parameter model. In practice the time of the hessian computation is always added, but even then the bestfit is (slightly) faster. Additionally, when running the Tutorial notebook with the bfgs optimizer and hessian I encountered no objective calls with nan values, whereas we have a few of those in the current (master) setup. It also seems that the guess position can be further away from the true position using bfgs+hessian.

pelssers:
Thanks. Yes, and we've lost the overhead from tracing the objective itself. Not tracing the objective also means that we can put more informative print statements there.
I'll merge.

PR #21, opened 2019-09-14
Generic source, conditional inference, padding bugfix

JelleAalbers:
This moves flamedisx's code around to more easily construct sources unrelated to liquid xenon TPCs. 
  * The `Source` class retains all generic functionality (fetch/gimme, populate cache, etc.)
 * A new class `LXeSource(Source)` contains the LXe-specific simulation, annotation, and model functions common to ER and NR.
 * `ERSource(LXeSource)` and `NRSource(LXeSource)` contain the ER/NR specific model defaults and functions. Note NR is no longer confusingly a subclass of ER.

(Besides this only minor things like adding some comments).


JelleAalbers:
Added a few more things:
 * Fixed a big bug :ant:: when computing the log_likelihood, we omitted padding events in all batches EXCEPT the last, rather than just the last. So if you had 41 events and a batch size of 10, you'd only condsider event 0, 10, 20, 30, 40 and spuriously add nine padding events (actually copies of event 0-8)... I hope this solves a few mysteries, e.g. considering stability of optimization with free rate_multipliers, and the rate_multiplier's guess.
 * Allow fixing parameters in the best-fit using `bestfit(fix=dict(bla=42))`. This conditional inference is essential to do profile likelihoods.
 * `summary`'s bestfit argument is now optional: if you just call `ll.summary()` it will do the bestfit for you. As with bestfit, you can pass a `fix` dictionary, which affects the errors.
 * inverse_hessian, log_likelihood and their sister functions now have an `omit_grads` option. This lets us skip unneeded gradients/Hessians in conditional inference.
 * `bestfit` now returns and `inverse_hessian` now accept `{parameter: value}` dictionaries, rather than a tensorflow tensor / numpy array.

Few things maybe of interest to @pelssers:
  * Maybe you told me this already, but it seems bfgs_minimize gives an inverse_hessian_estimate in its results, even if we do not start it with an initial inverse Hessian. Might be interesting to see if this is accurate?
  * I see LogLikelihood._log_likelihood_grad2_tf is not used. Maybe because the Hessian did not work with tf.function?

pelssers:
Thanks!
Great that you caught the bug in our event padding code!
It's nice that you introduced `LXeSource`. Having `NRSource` derive from `ERSouce` always looked confusing to me.

Interesting that the bfgs optimizer also returns the `inverse_hessian`. If this is indeed equal to the `inverse_hessian` at the bestfit than it will save us computing the `inverse_hessian` explicitly on the bestfit as we do now.
`_log_likelihood_grad2_tf` is not used. It's never selected in `log_likelihood`, if you pass `second_order=True` you will always get the `_log_likelihood_grad2`. The reason is indeed that the tracing doesn't seem to work with the second order derivatives, I tested this with `_log_likelihood_grad2_tf`. We could remove it.

flamedisx/er_nr_base.py
```diff
@@ -217,15 +645,11 @@ def rate_nq(self, nq_1d, data_tensor, ptensor):
...
-
     def _simulate_nq(self, energies):
         # OK to use None, simulator has set defaults
         work = self.gimme('work', data_tensor=None, ptensor=None, numpy_out=True)
         lindhard_l = self.gimme('lindhard_l',
                                 bonus_arg=energies,
                                 data_tensor=None, ptensor=None,
                                 numpy_out=True)
-        return stats.poisson.rvs(energies * lindhard_l / work)
+        return stats.poisson.rvs(energies * lindhard_l / work)
```
pelssers: Why does your editor always remove the newlines at the end of files?

JelleAalbers: I think it's just me :-)

tests/test_inference.py
```diff
@@ -151,7 +151,8 @@ def test_hessian(xes: fd.ERSource):
         data=xes.data)
 
     guess = lf.guess()
-    assert guess.shape == (2,)
+    assert len(guess) == 2
+    assert len(guess) == 2
```
pelssers: Duplicate line

JelleAalbers: I just wanted to be extra sure :-P

PR #22, opened 2019-09-19
Sr1 ER class

CristianAntochi:
I added the file for the Xenon1t SR1 with an SR1Source class and SR1ERSource class.
The model used is the one from the analysis paper (arXiv:1902.11297v2), with the correct yield maps. The ER model is not exactly complete and more features should be added soon, but it is workable and the key parameters can be fitted correctly.
The main difference from the model in the paper is the g1 parameter which needs a different default value for the model to fit the data i.e. g1 = 0.123 instead of g1 = 0.142.

Work to be done
The main things that need work are:
1) adding the functional time dependence for the electron lifetime
2) modifying the treatment of double pe fraction and reparametrize the model accordingly
3) add the SR1NRSource

side note
I modified also the setup.py file and corrected my name


flamedisx/x1t_sr1.py
```diff
@@ -0,0 +1,107 @@
...
+        return tf.clip_by_value(q2 * (tf.constant(1.,dtype=fd.float_type()) - tf.exp(-nq / q3_nq)),
+                                tf.constant(1e-4,dtype=fd.float_type()),
+                                float('inf'))
+    @staticmethod
+    def s2_acceptance(s2):
+        return tf.where((s2 < 500) | (s2 > 6000),
+                        tf.zeros_like(s2, dtype=fd.float_type()),
+                        tf.ones_like(s2, dtype=fd.float_type()))
+
+#TODO: add NRsource
```
pelssers: newline :)

CristianAntochi: done!

flamedisx/x1t_sr1.py
```diff
@@ -0,0 +1,107 @@
...
+    for x in ('XENON1T_s1_xyz_ly_kr83m-SR1_pax-664_fdc-adcorrtpf.json',
+              'XENON1T_s2_xy_ly_SR1_v2.2.json')]
+
+
+##
+# Flamedisx sources
+##
+
+
+@export
```
pelssers: Do we need @export here? I guess only on SR1ERSource and SR1NRSource or do we use SR1Source directly?

CristianAntochi: We might for development in notebooks for example for the NRSource. Don't we,  @JelleAalbers ?

pelssers: For developing we can just do `from flamedisx.x1t_sr1 import SR1Source` right? Or does the export mechanism prevent that somehow? I guess we don't want users to see `SR1Source` since its not fully specified.

JelleAalbers: Both approaches seem valid. I tend to lean towards exporting most things myself, but we can do either. Currently we're not very consistent in flamedisx (eg SR0Source is exported, as is SR0WIMPSource, but SR0ERSource is not..), so maybe it's easiest to just go through this once after both of your PRs are merged.

PR #24, opened 2019-09-23
Time dependent rate

pelssers:
This PR implements:

- Time dependent energy spectra for WIMPs.
- Dynamically setting new datasets for sources via `set_data`

### Time dependent energy spectra
The class `WIMPSource` is added which is derived from `NRSource`. The WIMP source uses the `rate_wimp_std` from the `wimprates` package to compute energy spectra for given WIMP settings (including WIMP mass, cross section and also time of detection).
WIMP settings can be set through class attributes or passed to init as `wimp_kwargs` dict.

The energy spectra are then used in simulation and data. Instead of calling `wimprates` for each event the energy spectra are interpolated in a given time range (the time range and number of reference values for the interpolation are class attributes).

Additionally the `SR0WIMPSource(SR0Source, fd.WIMPSource)` class can be used for SR0 settings (although the time range for WIMP interpolation is set larger then SR0 so that we can test on calibration data).

Since the energy spectra are precomputed and interpolated for performance reasons, the `energy_spectra` method is now a function of `i_batch` instead of time. In this way the energy spectra for a batch by looking up the relevant precomputed spectra.

When simulating events the energy time correlation is taken into account by drawing randomly from a 2D energy time histogram instead of each observable separately.  

### Source data via `set_data`
Instead of constructing a new Source object, the data a Source holds can be set via `set_data`. `batch_size` cannot be changed, but event padding is applied and Tensor caches are automatically rebuilt.


pelssers:
I've also added `set_data` to `LogLikelihood`. You can pass it the data in the same way as it's passed to the initializer. For likelihoods with multiple datasets you can also pass a subset to `set_data` to only update those datasets.
I've haven't changed anything to the initializer itself, so constructing `LogLikelihood` always requires data (contrary to `Source` which can be initialized without data for annotation and simulation purposes). @JelleAalbers Do you think we should support `data=None` for `LogLikelihood` as well?

Issue #23, opened 2019-09-23
Single column fix_truth 


pelssers:
When simulating events via `random_truth` we can fix the position and drift time of events via `fix_truth`. This should be extended so that `fix_truth` takes a dict fixing each key to value. Support keys to be any subset of observables and parameters.

pelssers:
Solved by #28

PR #28, opened 2019-10-04
Fixing observables in simulate

pelssers:
_This builds on #24 and #27 which need to be merged first (after which this branch might need to be rebased on master)_

This rewrites `random_truth` of `LXeSource` to solve issue #23.

When simulating events, random true values are drawn using `random_truth`. It was possible to fix the position of events by passing an event (in the form of a one-row pandas DataFrame) with `fix_truth`. This PR makes `fix_truth` also accept a dictionary whose keys can be any event observable and whose values are used for the simulated events instead of random values. This allows us to fix either the position, time or energy (or a combination thereof) when simulating events.
As a side effect of this more general implementation the `WIMPSource` now also works with an optional spatial response histogram.

Requirements on `fix_truth`:
 - a one-row pandas DataFrame that has at least the keys `['x', 'y', 'z', 'r', 'theta', 'event_time', 'drift_time']`. Exactly these variables will be fixed to the passed values during simulation.
 - a dictionary with keys: `['x', 'y', 'z']` and/or `['r', 'theta', 'z']` and/or `event_time` and/or `energy` to set either position, time, energy or any combination thereof.

When passing the position of events in Cartesian format, the cylindrical coordinates are automatically calculated as well and vice versa. When passing the position, the drift time is recalculated since it depends on the depth.

### DIFF
--- a/flamedisx/er_nr_base.py
+++ b/flamedisx/er_nr_base.py
@@ -179,19 +179,7 @@ class LXeSource(fd.Source):
     # Simulation
     ##
 
-    @staticmethod
-    def cart_to_pol(x, y):
-        return (x**2 + y**2)**0.5, np.arctan2(y, x)
-
-    @staticmethod
-    def pol_to_cart(r, theta):
-        return r * np.cos(theta), r * np.sin(theta)
-
     def random_truth(self, energies, fix_truth=None, **params):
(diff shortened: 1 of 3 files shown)


JelleAalbers:
Sorry, rebase fail

PR #31, opened 2019-10-09
WIMPSource bugfix and test

pelssers:
This fixes a bug introduced in PR #30 where `self.es` had been removed in too many places. This bug was not caught by our tests since `WIMPSource` is not tested against.

 * I've added a quick fix which solves the immediate problem. But we should rewrite the `WIMPSource.__init__` since it's not so well designed at the moment. I've opened issue #32  for this.

 * I've added `WIMPSource` to the pytest fixture in `test_source.py` so that it also runs our source tests.


pelssers:
Thanks for the rewrite, this makes the init a lot cleaner.

Issue #33, opened 2019-10-09
Batch size issue when using ColumnSource with other sources


CristianAntochi:
When using ColumnSource together with another source, when the LogLikelihood is initialised and then called, there is an error about different shape of the data tensors. When changing the batch size, the shape mismatch changes as well: if the batch size in the likelihood initialisation is 50 you would get a dimension 50 for the first source (in my case SR1NRSource) and dimension, lets say 77 for the ColumnSource. Viceversa if the batch size in the initialisation is set to 77, the first source will have dimension 77 and the second 50 (using the same data). 
It will occasionally work for one specific batch size, but it might be fortuitous and the result of a simultaneous fit cannot be trusted.

pelssers:
After looking into it with @CristianAntochi we think we identified the source of the problem.
The `data_tensor` is incorrectly reshaped in `ColumnSource.__init__`. https://github.com/JelleAalbers/flamedisx/blob/master/flamedisx/source.py#L84
The number of batches should be the first dimension, not the batch size.

Issue #29, opened 2019-10-08
Key error in extra_needed_columns for data


CristianAntochi:
When trying to pass a data frame with data with the added columns of 'x_observed' and 'y_observed' there is a key error in the initialisation of the LogLikelihood. 
Somehow the extra columns added to the data are either removed or not considered in the likelihood before looking up the values in the correction maps.

pelssers:
We looked into it and the issue is that the `x_observed` and `y_observed` columns are not simulated in `random_truth`. So even though data has the columns and they are kept when annotating, also random truth needs to be modified. So in the `SR1NRSource` you want to do something similar as was done in the SR0 Source:
https://github.com/JelleAalbers/flamedisx/blob/master/flamedisx/x1t_sr0.py#L82

PR #34, opened 2019-10-10
Update sr1 ER and NRSource, fix bugs and usage of ColumnSource

CristianAntochi:
In this pull request a few key features were fixed and updated:

- In the x1t_sr1.py the SR1Source was updated to take random_truth, which makes ERSource usable again
- The NRSource was added and takes also the extra needed columns of 'x_observed' and 'y_observed' from the correction maps #29 
- The use of ColumnSource was checked and a few bugs were fixed:
   - defaults were added to the initialisation of the source
   - the source can be initialised with data = None
   - an important bug was fixed in the shape of data given to the source, where the number of batches was given instead of the batch size #33 

Work to be done: 
Update the NRSource model to fit the AmBe data better and possibly update the defaults of it.

### DIFF
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -57,7 +57,7 @@ class ColumnSource(SourceBase):
     mu = 42.
 
     def __init__(self,
-                 data,
+                 data=None,
                  batch_size=10,
                  max_sigma=3,
                  data_is_annotated=False,
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -85,7 +85,7 @@ class ColumnSource(SourceBase):
         else:
             self._init_padding(batch_size, _skip_tf_init)
             self.data_tensor = fd.np_to_tf(self.data[self.column])
-            self.data_tensor = tf.reshape(self.data_tensor, (self.batch_size, -1, 1))
+            self.data_tensor = tf.reshape(self.data_tensor, ( -1,self.batch_size, 1))
 
     def differential_rate(self, data_tensor, **params):
         return data_tensor[:, 0]
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -76,7 +76,7 @@ class ColumnSource(SourceBase):
         :param fit_params: List of parameters to fit
         :param params: New defaults
         """
-        self.params = dict()
+        self.defaults = dict()
         self.data = data
 
         self._init_padding(batch_size, _skip_tf_init)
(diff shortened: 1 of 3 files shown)


flamedisx/x1t_sr1.py
```diff
@@ -29,12 +29,23 @@
...
 @export
-class SR1Source(fd.ERSource):
+class SR1Source:
     drift_velocity = 1.335 * 1e-4   # cm/ns
     extra_needed_columns = tuple(
         list(fd.ERSource.extra_needed_columns)
         + ['x_observed', 'y_observed'])
 
+    def random_truth(self, energies, fix_truth=None, **params):
+        d = super().random_truth(energies, fix_truth=fix_truth, **params)
```
pelssers: `energies` was renamed to `n_events` in #30 since we don't support arrays here any more. I'd rename it for consistency.

flamedisx/x1t_sr1.py
```diff
@@ -104,4 +115,10 @@ def s2_acceptance(s2):
                         tf.zeros_like(s2, dtype=fd.float_type()),
                         tf.ones_like(s2, dtype=fd.float_type()))
 
-#TODO: add NRsource
+class SR1NRSource(SR1Source, fd.NRSource):
+    extra_needed_columns = tuple(set(
+        list(SR1Source.extra_needed_columns) + 
+        list(fd.NRSource.extra_needed_columns)+
+        ['x_observed', 'y_observed']))
```
pelssers: I think `['x_observed', 'y_observed']` here is redundant since it's already in `SR1Source.extra_needed_columns`.

flamedisx/source.py
```diff
@@ -76,12 +76,16 @@ def __init__(self,
...
-        self.data_tensor = fd.np_to_tf(self.data[self.column])
-        self.data_tensor = tf.reshape(self.data_tensor, (self.batch_size, -1, 1))
+        if data is None:
+            # We're calling the source without data. Set the batch_size here
+            # since we can't pass it to set_data later
+            self.batch_size = batch_size
+        else:
+            self._init_padding(batch_size, _skip_tf_init)
+            self.data_tensor = fd.np_to_tf(self.data[self.column])
+            self.data_tensor = tf.reshape(self.data_tensor, ( -1,self.batch_size, 1))
```
pelssers: Thanks for fixing this reshape bug!

CristianAntochi:
Thanks for the comments. I addressed them!

Issue #26, opened 2019-10-01
Support LogLikelihood construction without data


pelssers:
For simulation purposes we want to support creating a `LogLikelihood` object with `data=None`. This would also mirror the implementation for `Source`.
We can already `set_data` on sources and likelihoods since PR #24 so only the initializer needs to be changed. Related to #16 as well.

pelssers:
Solved by #43

PR #43, opened 2019-10-25
Construct Likelihood without data

pelssers:
This PR implements construction of likelihoods without data to solve issue #26.

`LogLikelihood` can now be initialized with `data=None`, then call `set_data(data)` on your likelihood later to set dataset `data`. Take care to choose an appropriate `batch_size` in case you are dealing with small datasets (the number of events needs to be at least half the batch size for the padding to work). When passing very small datasets to the initializer (of either Source or LogLikelihood) `batch_size` is adjusted but this is of course not possible when `data=None`. I've added an assert for this in the `Source._init_padding` and tests for constructing single and multiple source likelihoods without data.

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -94,8 +94,8 @@ class LogLikelihood:
 
         # Create sources. Have to copy data, it's modified by set_data
         self.sources = {
-            d = data[self.d_for_s[sname]]
-            sname: sclass(data=None if d is None else d.copy(),
+            sname: sclass(data=None if data[self.d_for_s[sname]] is None \
+                            else data[self.d_for_s[sname]].copy(),
                           max_sigma=max_sigma,
                           fit_params=list(common_param_specs.keys()),
                           batch_size=batch_size)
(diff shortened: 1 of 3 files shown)


flamedisx/inference.py
```diff
@@ -90,7 +94,8 @@ def __init__(
 
         # Create sources. Have to copy data, it's modified by set_data
         self.sources = {
-            sname: sclass(data[self.d_for_s[sname]].copy(),
+            sname: sclass(data=None if data[self.d_for_s[sname]] is None \
```
JelleAalbers: Maybe you could look into the formatting here

pelssers: If only we had Python 3.8 :) 
```python
data=d if (d := data[self.d_for_s[sname]]) is None else d.copy()
```
but I'll remove the backslash for you.

pelssers:
Thanks, yes, you're right about the half-padding requirement. I only encountered it because our test case has only two events and we first constructed the likelihood without data resulting in a default `batch_size` value, normally you wouldn't see this behaviour. At least there is an assert for it now. Padding by repeating the same event would indeed be more flexible, I'll open an issue for it.

PR #48, opened 2019-10-31
Revert default to bfgs, add minuit stepsize

JelleAalbers:
This:
  * Changes the default optimizer back to bfgs. For running the tutorial on a CPU, this means the fits complete in 5-10 seconds rather than a minute.
  * Add a default stepsize of 0.1 * the guess to the minuit parameters. If we don't do this we get a warning and it will use 1 as the stepsize, at least for the rate multiplier (for which this is very large)
  * Make bestfit just return the bestfit dictionary without errors as default, so we don't need to do [0] every time. You can get the error dict by passing return_errors=True if you are using minuit. Alternatively we can look into returning some kind of FitResult namedtuple, with .errors and .result as attributes.
  * Adds explicit colab instructions to the tutorial.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -287,10 +287,10 @@ def test_bestfit_tf(xes):
     guess['er_rate_multiplier'] = xs[np.argmin(ys)]
     assert len(guess) == 2
 
-    bestfit = lf.bestfit(guess, optimizer=tfp.optimizer.bfgs_minimize, use_hessian=True)
-    assert isinstance(bestfit[0], dict)
-    assert len(bestfit[0]) == 2
-    assert bestfit[0]['er_rate_multiplier'].dtype == np.float32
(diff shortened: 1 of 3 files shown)


tests/test_inference.py
```diff
@@ -287,10 +287,10 @@ def test_bestfit_tf(xes):
     guess['er_rate_multiplier'] = xs[np.argmin(ys)]
     assert len(guess) == 2
```
CristianAntochi: `import tensorflow_probability as tfp` and`from iminuit import Minuit` are not used anymore in the tests. We can delete them.

CristianAntochi:
This looks good! Maybe we can think about the output of Minuit as well. We can use the full output of hesse like the covariance matrix and the correlation. But maybe we can do that in the future. Timewise the bfgs is certainly better

Issue #50, opened 2019-11-01
Likelihood initialisation with source specific parameters


CristianAntochi:
Right now the parameters in the likelihood init have to be common parameters to all the sources used (excluding the column source).
If we want to fit a full dataset with, say, ERSource and NRSource, the 2 of them will have source specific parameters besides the common parameters. Right now we don't support having source specific parameters and this should be a key feature that needs implementation.

CristianAntochi:
Solved by #51

Issue #47, opened 2019-10-31
ColumnSource simulator


pelssers:
@CristianAntochi just pointed out that ColumnSource has no `simulate` method. How would we implement one given that the differential rate of events in a ColumnSource are precomputed and passed in via data.

pelssers:
As @JelleAalbers said out during the review of #51 we could do this in the meantime:
>For columnsource perhaps we can add a default simulate that returns no events and emits a warning to tell the user they should provide the method?

pelssers:
Implemented @JelleAalbers suggestion in #54. SourceBase now has an 'abstract' simulate method. A test is added.

JelleAalbers:
Closing, since a generic ColumnSource simulate method is not meaningful / not defined. Users should provide this if using a ColumnSource. You added the warning, that's the best we can do.

We could someday add more advanced ColumnSource-like things, e.g. HistogramSource, which would have a simulate method.

PR #51, opened 2019-11-04
Different parameters to different sources in the initialisation of the likelihood

CristianAntochi:
As stated in issue #50 I have come across a problem when trying to use different sources in the likelihood. If the different sources use different parameters the likelihood initialisation would fail since it was designed to use the same sets of parameters for all the sources. 
With this pull request I have changed the way we filter the parameters and appropriately give them to the sources they are used in.
To do so I had to initialise the sources with `data = None` a first time, read the defaults and compare the defaults with the parameters given. If the parameters given are in the defaults of one source they are given to a second final initialisation of the source they are used in, this time with the data as well. 

This is probably not the most elegant way to do it, but it works and it makes the initialisation of the likelihood just a bit slower.

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -96,7 +96,7 @@ class LogLikelihood:
 
         # Determine default parameters for each source
         defaults_in_sources = {
-            sname : sclass.find_defaults()[2]
+            sname : sclass().find_defaults()[2]
             for sname, sclass in self.sources.items()}
 
         # Create sources. Have to copy data, it's modified by set_data
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -93,14 +93,22 @@ class LogLikelihood:
             if sn not in self.sources:
                 raise ValueError(f"Can't free rate of unknown source {sn}")
             param_defaults[sn + '_rate_multiplier'] = 1.
-
+        
+        sources_null_init = {
+            sname: sclass(data=None)
+            for sname, sclass in self.sources.items()}
+        defaults_in_sources = {
+            sname : sclass.defaults
+            for sname, sclass in sources_null_init.items()}
+            
         # Create sources. Have to copy data, it's modified by set_data
         self.sources = {
             sname: sclass(data=(None
                                 if data[self.d_for_s[sname]] is None
                                 else data[self.d_for_s[sname]].copy()),
                           max_sigma=max_sigma,
-                          fit_params=list(common_param_specs.keys()),
+                          fit_params=list(k for k in common_param_specs.keys() 
+                                          if k in defaults_in_sources[sname].keys()),
                           batch_size=batch_size)
             for sname, sclass in self.sources.items()}
         del data  # use data from sources (which is now annotated)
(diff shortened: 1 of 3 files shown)


CristianAntochi:
Added a small fix in the simulate method to work with different parameters for different sources and an if statement to avoid the simulation of ColumnSource... For now.

flamedisx/inference.py
```diff
@@ -176,24 +185,26 @@ def simulate(self, rate_multipliers=None, fix_truth=None, **params):
...
-            # Simulate this many events from source
-            n_to_sim = np.random.poisson(mu)
-            if n_to_sim == 0:
-                continue
-            d = s.simulate(n_to_sim,
-                           fix_truth=fix_truth,
-                           **params)
-            d['source'] = sname
-            ds.append(d)
+            if s.defaults:
```
pelssers: This is just to ignore ColumnSource for now because of #47 right? I'd put a TODO here to make sure we don't forget to remove this later.

flamedisx/source.py
```diff
@@ -215,8 +215,6 @@ def set_defaults(self, **params):
             if k in self.defaults:
                 self.defaults[k] = tf.convert_to_tensor(
                     v, dtype=fd.float_type())
-            else:
-                raise ValueError(f"Key {k} not in defaults")
```
pelssers: Is this still needed now that we pass params filtered by `defaults_in_sources[sname]`?

flamedisx/inference.py
```diff
@@ -176,24 +185,26 @@ def simulate(self, rate_multipliers=None, fix_truth=None, **params):
...
+                # mean number of events to simulate, rate mult times mu source
+                mu = rm * self.mu_itps[sname](**self._filter_source_kwargs(params,
+                                                                           sname))
+                # Simulate this many events from source
+                n_to_sim = np.random.poisson(mu)
+                if n_to_sim == 0:
+                    continue
+                d = s.simulate(n_to_sim,
+                               fix_truth=fix_truth,
+                               **self._filter_source_kwargs(params,
```
pelssers: Good that you caught this as well

flamedisx/inference.py
CristianAntochi: Yes, indeed!
I will do that!

flamedisx/source.py
CristianAntochi: It is needed for the simulation as it takes all the fit params without distinction; the if checks if it's in the defaults and if not doesn't do anything. with the error there, if the parameter is not in the defaults for the specific source it will raise an error even if the parameter is not given then to the source.

pelssers: Ok, thanks

pelssers:
I've implemented the suggestion above to avoid initializing the sources twice. It moves the discovery of the function parameters to a classmethod which `LogLikelihood` uses to get the defaults without initializing the source.

pelssers:
When `find_default` is a `classmethod` the tests fail. I'm not exactly sure why but have removed the decorator for now.

CristianAntochi:
Looks nice! Thanks!

pelssers:
I've fixed the function inspection so now `find_defaults` works again for class methods.

pelssers:
Agreed, I'll add you comment to issue #47

Issue #45, opened 2019-10-29
More flexible event padding


pelssers:
As encountered in #43 our event padding code only works when `n_events * 2 >= batch_size`. In practise we don't run into this because we set the batch size appropriately. However, this requirement can be removed by padding with the same event repeatedly. Additionally we could pad with some special placeholder event as long as it doesn't cause issues in the differential rate computation.

pelssers:
Actually, we do run into this when simulating with small rate multipliers. The case for zero events is fixed in #46 but this still limits us to using `batch_size=2` if there is a chance we only simulate one event.

pelssers:
Event padding on single event data is implemented in #54

pelssers:
Event padding is now flexible, fixing this issue. I don't think we'll want to implement different batch sizes for different sources in the likelihood, but could open a new 'enhancement' issue for this. Closing.

JelleAalbers:
I guess we'll want to support the 0-event case eventually as well; it might come up during testing with toy MCs. But for this we can just bypass differential_rate altogether, we don't need to have an 'all padding' batch that is then discarded.

pelssers:
The 0-event case is already supported regarding the simulator in #46, I don't think we need to support `set_data` with 0 events since that is clearly not what `set_data` is for.
I did see an assert in `source.py` that still needs to be removed btw.

JelleAalbers:
OK, good that the simulator already supports this. Observing no events is a legitimate experimental outcome, if set_data doesn't support this we would have to crash when this ever happens in a toy MC. Or maybe we could catch this case in LogLikelihood so the source is never involved.

PR #54, opened 2019-11-18
Refactor, minimizer and interval estimation

pelssers:
This restructures flamedisx, adds the SciPy optimizer and adds interval estimation.

### Minimizer backends and options ###

- Rate parameters now have bounds `(1e-9, None)` so the minimizer will not search in unphysical negative rate parameter space.
 -  `SciPy.optimize.minimize` is added as optimizer backend.

### Interval estimation ###

- A `one_parameter_interval` method is added to `LogLikelihood` with its main implementation in `inference.py`.
- `one_parameter_interval` computes upper, lower limits or central intervals for a parameter of the likelihood at a given confidence level.
- Critical values are determined according to Wilk's theorem by default (asymptotic case) but a user defined critical value curve can be supplied (together with it's gradient, this is mandatory) for the non-asymptotic case.
- The function minimized is the square of the loglikelihood ratio as function of the parameter minus the critical value as function of the parameter.
- More tests are needed for this, only basic test in place.

### Refactor ###

- `inference.py` is renamed to `likelihood.py` since it mainly contains the `LogLikelihood` class.
- All implementation related to minimization of the likelihood function and other inference are moved to a new `inference.py` file. The `LogLikelihood` class does keep a `bestfit` (and `one_parameter_interval`) method for convenience since it's the main class that users should interact with.
- Several minimizers supported through common `Objective` implementation. Since we had two main minimizer backends (either optimizers from TensorFlow probability and the Minuit optimizer from iminuit) we had several sets of 'objective' functions and their gradients cluttering the `LogLikelihood` class. The `Objective` class in `inference.py` now defines the common functionality for `bestfit`. For each minimizer backend a small subclass with the details of the backend's `minimze` call is all that is required to add another backend. `SciPy.optimize.minimize` is added as the third backend. It should also be easy to add for example zfit in this way.
- Calls to the objective function (returning the likelihood and gradient) are memoized to increase speed. SciPy and Minuit require separate func and grad functions while we compute the likelihood and gradient in one.
 - We could clean up the arguments passed to optimizers more and support more optimizer features, such as passing a hessian to the SciPy backend which is currently not implemented.
 - The `Objective` class has a set of tests that use a `MockObjective` to test just the optimizer performance independent of the likelihood. The mock likelihood is just a parabola for each parameter and has some basic tests to see if the minimizer recovers the true minimum.

### SourceBase.simulate ###

- SourceBase now has a simulate method that returns an empty DataFrame. This 'abstract' method ensures that LogLikelihood.simulate doesn't crash on sources without simulate defined, the workaround that was added in #51 is removed. All sources should override `SourceBase.simulate`, if they don't the simulation for that source is skipped and a warning is printed. We should raise an error, but not until #47 is fixed.

### Flexible event padding ###

- Event padding now works on single event data, repeating the first event n times instead of taking the first n events. See #45


flamedisx/source.py
```diff
@@ -63,7 +75,8 @@ def _init_padding(self, batch_size, _skip_tf_init):
             # padded events are clipped when summing likelihood terms
             self.n_padding = self.n_batches * self.batch_size - len(self.data)
             if self.n_padding > 0:
-                df_pad = self.data.iloc[:self.n_padding, :]
+                # Repeat first event n_padding times and concat to rest of data
+                df_pad = self.data.iloc[np.full(self.n_padding, 0)]
```
JelleAalbers: Can use np.zeros here too

pelssers: Done

PR #58, opened 2019-11-21
Simulator checks and timestamps

pelssers:
Larger changes and improvements:

 - Ensure that `Source.differential_rate` only needs to be traced once during the lifetime of a Source object.
   - Tracing is now defined in the source initialization since the input signature is known.
   - We use `tf.function` to graph the diff rate function and supply an input signature for it. This means that if only the values (and not the shape and dtype) of `data_tensor` and `ptensor` change we don't need to retrace. However we were actually retracing on every `set_data` call which is now fixed. This means that we also had to merge `energy_tensor` and `spatial_rate_tensor` into `data_tensor` so that all is passed through `_differential_rate`.
- Remove all tracing in `log_likelihood`
  - Most of the performance gain from using `tf.function` is due to `differential_rate`, also tracing `log_likelihood` doesn't add more speed but instead introduces overhead from tracing. `log_likelihood` also needs to be traced every `set_data` (which also retraces diff rate unnecessarily). Removing all tracing in `LogLikelihood` speeds up toyMCs from about 24 s per toy (including simulation, bestfit and interval estimation) to about 18 s.
 - Reduced memory usage for toyMCs
   - Repeatedly tracing functions using `tf.function` takes up a lot of RAM. Before we could not run more than 400 toysMCs in one runtime. Tracing only once (on diff rate) fixes this.

Smaller changes and improvements:

- Added SR1WIMPSource class to `x1t_sr1.py`, exporting `SR1NRSource` and `SR1WIMPSource` but not `SR1Source` to be consistent with `x1t_sr0.py`
- Made `LogLikelihood.simulate` more robust:
  - Added a check that checks if all sources that were simulated returned DataFrames with the same columns, this is important since columns not in all sources will be filled with NaN values that might lead to problems later on.
  - Skip sources that return 0 events (either because no events are left due to efficiencies or because they don't have simulate implemented).
  - Ensure that pd.concat doesn't crash if all sources returned 0 events (which is possible with small rate_multipliers) by adding an empty DataFrame to the list.
- ~~One case where different sources returned dataframes with different columns was when combining ERSource with WIMPSource. The ERSource does not have the J2000 't' times and as a result after `LogLikelihood.simulate` these would be NaN. This behaviour is now caught by the above check. Additionally generating the J2000 times is moved up to `LXeSource` such that we now can combine all Sources derived from `LXeSource`.~~
  - It turns out `add_extra_columns` in `SR1Source` didn't call its super() so was overwriting the method from `LXeSource`. This is why data was not annotated with the J2000 timestamps for WIMP sources and I fixed it by computing J2000 timestamps for all sources.
Now that `add_extra_columns` works correctly I've restored the original implementation.
- Relaxed the requirements on `tensorflow` and `tensorflow_probability` in requirements, adjusted `setup.py` accordingly.
- I've also cleaned up the conversion of J2000 timestamps back to event times (since we now know it's simply scaling unix time).
- pinned required wimprates version to >= 0.3.1 since the j2000 api has changed
- added j2000_to_event_time conversion in utils (and added test for it)
- Tried Py3.8 build but TF2 not yet supported, so keep at Py3.7 for now

### DIFF
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -176,10 +176,8 @@ class LogLikelihood:
         if rate_multipliers is None:
             rate_multipliers = dict()
 
-        # Collect Source event DFs in ds, adding empty DataFrame ensures
-        # pd.concat doesn't fail if n_to_sim is 0 for all sources or all
-        # sources return 0 events
-        ds = [pd.DataFrame()]
+        # Collect Source event DFs in ds
+        ds = []
         for sname, s in self.sources.items():
             rmname = sname + '_rate_multiplier'
             if rmname in rate_multipliers:
@@ -207,13 +205,16 @@ class LogLikelihood:
 
         # Check if each source returned the same set of columns since
         # columns outside the intersection will be filled with NaN values
-        if len(ds) > 2:
-            col_sets = set([frozenset(df.columns) for df in ds[1:]])
+        if len(ds) > 1:
+            col_sets = set([frozenset(df.columns) for df in ds])
             assert len(col_sets) == 1, \
                 f"Sources did not all simulate the same columns: {col_sets}"
 
-        # Concatenate results and shuffle them
-        return pd.concat(ds, sort=False).sample(frac=1).reset_index(drop=True)
+        # Concatenate results and shuffle them.
+        # Adding empty DataFrame ensures pd.concat doesn't fail if
+        # n_to_sim is 0 for all sources or all sources return 0 events
+        ds = pd.concat([pd.DataFrame()] + ds, sort=False)
+        return ds.sample(frac=1).reset_index(drop=True)
 
     def __call__(self, **kwargs):
         assert 'second_order' not in kwargs, 'Roep gewoon log_likelihood aan'
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -193,7 +193,10 @@ class LogLikelihood:
         if rate_multipliers is None:
             rate_multipliers = dict()
 
-        ds = []
+        # Collect Source event DFs in ds, adding empty DataFrame ensures
+        # pd.concat doesn't fail if n_to_sim is 0 for all sources or all
+        # sources return 0 events
+        ds = [pd.DataFrame()]
         for sname, s in self.sources.items():
             rmname = sname + '_rate_multiplier'
             if rmname in rate_multipliers:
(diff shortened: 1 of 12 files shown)


pelssers:
The Travis build time just tripled, I think it's due to the J2000 timestamps that are moved up to LXeSource, the conversion from `event_time` to `t` seems to be very slow.

pelssers:
I've vectorized wimprates J2000 computation: https://github.com/JelleAalbers/wimprates/pull/5
It can now take a numpy array of `event_times`.

pelssers:
I'm trying to see if it really is the J2000 timestamp computation which is the cause of the long build time, running the tests locally I get these results:

branch | pytest runtime [s]
-------- |--------
master | 332
cleanup | 941
cleanup without moving the j2000 comp to LXeSource | 339
cleanup with new j2000 implementation | 322

So it seems that it is indeed the J2000 computation and that the new implementation will fix build times again.

pelssers:
With the release of wimprates v0.3.2 (which has https://github.com/JelleAalbers/wimprates/pull/5) we can now use the vectorized j2000 computation. I've added a commit which implements the api change and removed the list comprehension.

pelssers:
It turns out `add_extra_columns` in `SR1Source` didn't call its super() so was overwriting the method from `LXeSource`. This is why data was not annotated with the J2000 timestamps for WIMP sources and I fixed it by computing J2000 timestamps for all sources.
Now that `add_extra_columns` works correctly I've restored the original implementation. I've also cleaned up the conversion of J2000 timestamps back to event times (since we now know it's simply scaling unix time).

flamedisx/likelihood.py
```diff
@@ -181,19 +169,17 @@ def set_data(self,
...
         rate_multipliers and fix_truth.
         """
         if rate_multipliers is None:
             rate_multipliers = dict()
 
-        ds = []
+        # Collect Source event DFs in ds, adding empty DataFrame ensures
+        # pd.concat doesn't fail if n_to_sim is 0 for all sources or all
+        # sources return 0 events
+        ds = [pd.DataFrame()]
```
JelleAalbers: Maybe you can add this right before the concatenate, so the code below is easier to understand (`len(ds) > 2` -> `len(ds) > 1`, `ds[1:]` -> `ds`). Or just skip the concatenate if not len(ds).

JelleAalbers:
I added a few commits, mostly to enable tracing the source on initialization (even when the data is not yet provided). The per-batch data tensor shape is now known and fixed after init. 

We might consider removing set_data again in the future... but now without requiring data on init, which led to all the weird solutions with classmethods from before. This would be a big refactor though, probably not worth the trouble now (or maybe at all).

pelssers:
Thanks, I see you now explicitly compute which data columns should be cached instead of caching every number-like column as we did before, this is much cleaner.
Also the extra_needed_columns are now used as they were intended.
Shame about the StaticVocabularyTables though :)

I'll update the PR text with all the new changes, maybe we can sit together to review all this.

PR #59, opened 2019-11-28
Optimizer fixes and limit guesses

JelleAalbers:
This adds:
  * Let minuit use bounds for rate multipliers by default. Scipy already did this; tensorflow-probability's optimizers do not support bounds and should not be used for WIMP limit setting currently (unless you like your WIMPs negative...)
  * Fix the default optimizer tolerances so that for all optimizers we support, to the best of my knowledge, llr_tolerance controls the absolute distance in -2 log likelihood between the solution and the true minimum. This is quite tricky and it's likely there are still some quite different behaviours, considering that each optimizer has a different algorithm with a different way of estimating the remaining distance to the minimum. Theit tolerance options are also not very clearly/consistently documented.
  * Change one_parameter_interval to no longer require a guess: we can make a reasonably good guess from the Hessian (or for boundary solutions, the gradient). Of course you can still provide a guess, though now you specify it via a float (or tuple of floats for central intervals) instead of a dictionary. 

Minor bonus changes:
  * Fix a crash related to kwarg handling in _get_rate_mult. I had problems passing the `rate_multipliers` argument to simulate. Is this argument just a relic that can be removed? We can pass rate multipliers via the normal kwargs.
  * Let one_parameter_interval take kwargs
  * Remove the check that all sources must produce the same columns. With this check, I cannot simulate any combined ER + WIMP dataset (unless the WIMP rate multiplier is zero, of course).

We should probably refactor a bit, e.g. create a separate method for the limit guess.


JelleAalbers:
Bonus change: ColumnSource now inherits from Source, and Sourcebase has been removed. This simplifies the code and avoids some missing method errors.

This does mean that ColumnSource will do batching and padding, which strictly speaking it doesn't need to. But this takes no extra logic and hardly any computation time. ColumnSource's random_truth (and therefore its simulator) always returns an empty dataframe (see #47), unless the user overrides this with a custom simulator.

Issue #53, opened 2019-11-17
Old data sticks around in autograph underworld after Likelihood.set_data


JelleAalbers:
When changing datasets with Likelihood.set_data (introduced in #24), autographed functions from tensorflow will continue to refer to the old dataset. This is most visible when you set a dataset with a larger number of batches: then likelihood evaluation with autograph=False works, but with autograph=True it crashes due to an index out of bounds error that can be traced back to here https://github.com/FlamTeam/flamedisx/blob/592562d4654cf77d69d55b91b145a2f4bb36f02d/flamedisx/inference.py#L351

There is actually a comment here https://github.com/FlamTeam/flamedisx/blob/592562d4654cf77d69d55b91b145a2f4bb36f02d/flamedisx/inference.py#L169 that says a retrace is needed (but not done) if the number of batches or padding changes -- which would be always, unless the number of events is exactly equal. In fact, even in that case, a retrace is needed (but not done) with the current implementation.

Likelihood, unlike Source, does not pass the data tensors as arguments to the traced functions. Instead, we get the data tensor from the state of the source (Source.data_tensor). Tensorflow's autograph references the tensor under Source.data_tensor at the moment of tracing, and does not trace the dynamic access to the python attribute of Source.data_tensor.

We should access Source.data_tensor in the python layer, and pass it as an argument to the tensorflow layer, along (I guess) with other needed info such as the amount of padding.

For the moment just not using Likelihood.set_data, or being religious about autograph=False if you do, are the only workarounds I can think of.

pelssers:
Interesting, what happens when you turn off autograph for `LogLikelihood._log_likelihood`, but keep it on for `Source.differential_rate`. My guess is that is works correctly that way. Since we get the main performance benefit from the tracing of `differential_rate` just turning off the outer `log_likelihood` tracing might be an easy fix until we rewrite `set_data` to trigger the retrace.

pelssers:
Implemented in #55 retracing the likelihood is now done after each call to `LogLikelihood.set_data` so no old data sticks around any more. This solves the bug, but performance has to be evaluated. As stated above, we could implement `_log_likelihood` similar to `Source.differential_rate`.

PR #61, opened 2020-01-24
Trace log likelihood, then no more tensorflow

JelleAalbers:
This:
  * Enables autograph mode for the `_log_likelihood` function, which computes the likelihood of a single batch and dataset (for all sources in that dataset).
  * Avoids using tensorflow in higher-level functions, i.e. convert to numpy immediately after `_log_likelihood`.
  * Removes the test for the tensorflow-probability minimizer. 

The main reason for this change is performance: eager tensorflow has a much higher per-computation overhead  than autograph tensorflow or numpy, but the computations after _log_likelihood are all small and on tiny bits of data -- both the sum over events and that over sources have already happened at that point.

For simple likelihoods made of ColumnSources, i.e. where the likelihood per event is precomputed, the eager-mode computation made up the vast majority of the computational work. For a simple test, I computed 11 limits of different CLs with a ColumnSource likelihood on my laptop CPU:
  * This took ~89 seconds in current master and ~2.8 seconds after this change (~30x speedup).
  * If guess sigma_guess are not passed in, these have to be derived from the Hessian, which is still an eager-mode computation (it still gives NaNs in autograph mode). Then the test takes ~94 seconds in the current master and ~4.7 seconds after this change (still a ~20x speedup).

To avoid #53, we pass the data tensors and batching info as function arguments. This means the likelihood does not have to be retraced after each `set_data`. @pelssers already implemented a test to check that this, in fact, works.

The data tensors are concatenated (along the second axis, i.e. columns_to_cache) into one super data tensor for each dataset. This avoids having to pass a variable-length list into the autograph functions, which would probably trigger retracing.

The tensorflow-probability minimizer was not used in practice since it did not support bounds; it _might_ still work after this change (at least it gets to OptimizerFailure) but since we now convert to numpy before hitting the optimizer there seems little point in converting back.

### DIFF
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -236,7 +236,7 @@ class LogLikelihood:
 
     def __call__(self, **kwargs):
         assert 'second_order' not in kwargs, 'Roep gewoon log_likelihood aan'
-        return self.log_likelihood(second_order=False, **kwargs)[0].numpy()
+        return self.log_likelihood(second_order=False, **kwargs)[0]
 
     def log_likelihood(self, second_order=False,
                        omit_grads=tuple(), **kwargs):
@@ -249,24 +249,26 @@ class LogLikelihood:
 
         params = self.prepare_params(kwargs)
         n_grads = len(self.param_defaults) - len(omit_grads)
-        ll = tf.constant(0., dtype=fd.float_type())
-        llgrad = tf.zeros(n_grads, dtype=fd.float_type())
-        llgrad2 = tf.zeros((n_grads, n_grads), dtype=fd.float_type())
+        ll = 0.
+        llgrad = np.zeros(n_grads, dtype=np.float32)
+        llgrad2 = np.zeros((n_grads, n_grads), dtype=np.float32)
 
-        for dset_index, dsetname in enumerate(self.dsetnames):
-            n_batches = self.batch_info[dset_index, 0]
+        for dsetname in self.dsetnames:
+            # Getting this from the batch_info tensor is much slower
+            n_batches = self.sources[self.sources_in_dset[dsetname][0]].n_batches
 
-            for i_batch in tf.range(n_batches, dtype=fd.int_type()):
-                v = f(i_batch,
+            for i_batch in range(n_batches):
+                # Iterating over tf.range seems much slower!
+                v = f(tf.constant(i_batch, dtype=fd.int_type()),
                       dsetname=dsetname,
                       data_tensor=self.data_tensors[dsetname],
                       batch_info=self.batch_info,
                       omit_grads=omit_grads,
                       **params)
-                ll += v[0]
-                llgrad += v[1]
+                ll += v[0].numpy()
+                llgrad += v[1].numpy()
                 if second_order:
-                    llgrad2 += v[2]
+                    llgrad2 += v[2].numpy()
 
         if second_order:
             return ll, llgrad, llgrad2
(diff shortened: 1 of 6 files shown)


flamedisx/likelihood.py
```diff
@@ -149,21 +155,54 @@ def set_data(self,
...
         if isinstance(data, pd.DataFrame):
-            # Only one dataset
             assert len(self.dsetnames) == 1, \
-                "You passed a DataFrame but there are multiple datasets"
+                "You passed one DataFrame but there are multiple datasets"
             data = {DEFAULT_DSETNAME: data}
 
+        if any([d is None for d in data.values()]):
+            # At least one dataset is None => reset everything!
+            # TODO: should this give a warning?
```
pelssers: Yes, also what does reset mean here. If all datasets are set to None we can only use the Source for simulation but a user doesn't first need to reset the Source to do that, so why support it. I guess it's less prone to subtle bugs if all data is set to None instead of one dataset, so without adding a warning this makes sense.

flamedisx/likelihood.py
```diff
@@ -93,17 +109,14 @@ def __init__(
             sname: sclass.find_defaults()[2]
             for sname, sclass in self.sources.items()}
 
-        # Create sources. Have to copy data, it's modified by Source.set_data
+        # Create sources
         self.sources = {
-            sname: sclass(data=(None
-                                if data[self.d_for_s[sname]] is None
-                                else data[self.d_for_s[sname]].copy()),
+            sname: sclass(data=None,
```
pelssers: This means we don't set the batch_size based on data size any more.
In source when data is not None we do: `self.batch_size = min(batch_size, len(data))`
Not sure if still needed.

flamedisx/likelihood.py
```diff
@@ -116,15 +129,6 @@ def __init__(
...
 
-        # Store n_batches, batch_size and n_padding for all datasets
-        self.n_batches = dict()
-        self.batch_size = dict()
-        self.n_padding = dict()
-        for sname, s in self.sources.items():
-            self.n_batches[self.d_for_s[sname]] = s.n_batches
-            self.batch_size[self.d_for_s[sname]] = s.batch_size
-            self.n_padding[self.d_for_s[sname]] = s.n_padding
-
```
pelssers: Nice cleanup

flamedisx/likelihood.py
```diff
@@ -263,64 +310,74 @@ def _param_i(self, pname):
...
-                continue
+        drs = tf.zeros((batch_size,), dtype=fd.float_type())
+        for source_i, sname in enumerate(self.sources_in_dset[dsetname]):
+            s = self.sources[sname]
             rate_mult = self._get_rate_mult(sname, params)
-            dr = s.differential_rate(s.data_tensor[i_batch],
-                                     autograph=autograph,
-                                     **self._filter_source_kwargs(params, sname))
+
+            col_start, col_stop = self.column_indices[dsetname][source_i]
```
pelssers: `column_indices` is constant for the lifetime of the Source object right?

flamedisx/likelihood.py
```diff
@@ -263,64 +310,74 @@ def _param_i(self, pname):
...
         for sname, s in self.sources.items():
-            if self.d_for_s[sname] != dsetname:
+            if self.dset_for_source[sname] != dsetname:
                 continue
             mu += (self._get_rate_mult(sname, kwargs)
                    * self.mu_itps[sname](**self._filter_source_kwargs(kwargs, sname)))
         return mu
 
-    def _log_likelihood(self, i_batch, dsetname, autograph,
+    @tf.function
```
pelssers: We could add a `input_signature` similar to how we do it in Source.

flamedisx/likelihood.py
JelleAalbers: Yes, sources can't change their mind about which columns they use

flamedisx/likelihood.py
JelleAalbers: Good point, added a batch size check in the likelihood init

PR #64, opened 2020-02-04
FV specification

pelssers:
This changes the specification of the TPC dimensions in `LXeSource` to use a simple cylindrical fiducial volume instead. By default the behaviour is the same since the whole TPC is set a FV. Using `fv_radius` the maximum radius of events is set. `fv_high` and `fv_low` set the maximum and minimum depth respectively.
Setting the old `tpc_radius` and `tpc_length` parameters now raises a `DeprecationWarning`

An unnecessary reshape in `add_extra_columns` has been removed.

S1,S2 placeholder values have been removed from `random_truth_observables`, they are no longer needed for `set_data` to work. This was originally needed for bounds computation.

Lastly when drawing `event_time` values (which are O(10^18) integers) don't convert them to floats to preserve precision (they will eventually end up in the data tensor as float types anyway, but keep them as integers for as long as possible (and for simulate)).

### DIFF
--- a/flamedisx/er_nr_base.py
+++ b/flamedisx/er_nr_base.py
@@ -95,9 +95,9 @@ class LXeSource(fd.Source):
         super().add_extra_columns(d)
         if self.spatial_rate_hist is not None:
             # Setup tensor of histogram for lookup
-            positions = self.data[self.spatial_rate_hist_dims].values.T
+            positions = d[self.spatial_rate_hist_dims].values.T
             v = self.spatial_rate_hist.lookup(*positions)
-            d['spatial_rate_multiplier'] = v.reshape([self.n_batches, -1])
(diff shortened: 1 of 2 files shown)


pelssers:
I've added a commit to implement the spatial rate pdf as well as the spatial rate events per bin histogram. This is needed for the lookup of the spatial rate multipliers. This means also the bin volumes have to be set when making the source.

pelssers:
The `ER_spatial` tests actually weren't running since we renamed `spatial_hist` to `spatial_rate_hist`, fixed now.

pelssers:
The `spatial_rate_pdf` histogram is now automatically scaled correctly so that we don't have to modify `mu_before_efficiencies` or the differential rates.

PR #69, opened 2020-04-09
Add sanity checks on the input data and guess

JelleAalbers:
This lets flamedisx check that the likelihood has a finite value at the guess, and checks the inferred S1 and S2 acceptances and detection efficiencies are positive for all events. A common case where the latter would fail is if you forget to apply cuts you told flamedisx about.

Curiously, the trust-constr optimizer throws some ValueError about dimension mismatch for the Hessian in case the guess's likelihood is nonfinite. Not sure what this is about, but I hope we won't see this is again with this check.

I introduced a new `_check_data` function in source, since the check on the S1 acceptance cannot happen directly in annotate (for simulated data, annotate is first called without S1 and S2 values defined ), and so we can also check all required columns are present in the data before producing KeyErrors in populate tensor cache.

### DIFF
--- a/flamedisx/er_nr_base.py
+++ b/flamedisx/er_nr_base.py
@@ -75,6 +75,13 @@ class LXeSource(fd.Source):
     spatial_rate_hist = None
     spatial_rate_bin_volumes = None
 
+    # Whether to check efficiencies and acceptances are positive at
+    # the observed events.
+    # This is recommended, but you'll have to turn it off if your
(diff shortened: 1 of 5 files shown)


JelleAalbers:
Turns out we actually created out-of-bounds S1s in one of our tests: https://github.com/FlamTeam/flamedisx/blob/master/tests/test_source.py#L278...

PR #70, opened 2020-04-25
Nonlinconstr

plt109:
Limit calculation using `trust-constr` using new objective function of y = -x with `scipy.optimize.NonlinearConstraint`. `NonlinearIntervalObjective` class inherits from a single `NonlinearObjective` class which inherits from `Objective`.

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -8,7 +8,6 @@ from scipy import optimize as scipy_optimize
 import tensorflow as tf
 import tensorflow_probability as tfp
 
-import pdb
 from scipy.optimize import NonlinearConstraint
 
 export, __all__ = fd.exporter()
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -694,25 +694,27 @@ class NonlinearIntervalObjective(IntervalObjective, ScipyObjective):
     def _minimize(self):
         kwargs = self._scipy_minizer_options()
         kwargs['method'] = 'trust-constr'
-        # The default trust radius of 1 seems too large for some problems,
-        # it immediately jumps to values giving NaN results in the constraint,
-        # which this optimizer cannot handle
-        kwargs['options'].setdefault('initial_tr_radius', 0.1)
 
         constraint = NonlinearConstraint(
             fun=self.fun,
-            # The constraint is self.fun == 0
-            # TODO: does having the same bounds give numerical instability?
+            # The constraint is self.fun == 0, so we should set these equal
+            # (NonlinearConstraint docstring confirms)
             lb=0,
             ub=0,
             jac=self.grad,
             hess=self.hess_constraint if self.use_hessian else None)
 
+        n = len(self.arg_names)
+        tilt_grad = np.zeros(n)
+        tilt_grad[self.arg_names.index(self.target_parameter)] = -self.direction
+
         return scipy_optimize.minimize(
             fun=self.tilt_fun,
-            jac=self.grad,
+            # Objective is linear
+            jac=lambda x: tilt_grad,
+            hess=lambda x: np.zeros((n, n)),
             # The minimizer operates in scaled coordinates
-            x0=np.ones(len(self.arg_names)),
+            x0=np.ones(n),
             constraints=constraint,
             **kwargs)
 
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -354,7 +354,7 @@ class ScipyObjective(Objective):
         # TODO implement optimizer methods to use the Hessian,
         # see https://github.com/FlamTeam/flamedisx/pull/60#discussion_r354832569
 
-        kwargs: ty.Dict[str, ty.Any] = self.optimizer_kwargs
+        kwargs: ty.Dict[str, ty.Any] = self._scipy_minizer_options()
 
         if self.use_hessian:
             # Of all scipy-optimize methods, only trust-constr takes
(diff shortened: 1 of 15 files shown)


flamedisx/inference.py
```diff
@@ -8,6 +8,8 @@
 import tensorflow as tf
 import tensorflow_probability as tfp
 
+import pdb as pdb
```
pelssers, line 11: Unused import

flamedisx/inference.py
```diff
@@ -263,8 +265,10 @@ def _minimize(self):
             # Of all scipy-optimize methods, only trust-constr takes
             # both a Hessian and bounds argument.
             kwargs.setdefault('method', 'trust-constr')
+            print('Using scipy trust-constr')
```
pelssers, line 268: If you've checked it's passing the argument correctly remove the print (also on line 271)

flamedisx/inference.py
```diff
@@ -391,12 +396,67 @@ def parse_result(self, result: Minuit):
...
+    #def fishPath(self, a, b):
+    #    self._callbackbag.append(a)
+
+    def _inner_fun_and_grad(self, params):
+        # lol who exactly is this Objective._inner_fun_and_grad(self, params)?
+        m2ll, grad, hess = Objective._inner_fun_and_grad(self, params) 
+        return m2ll - self.m2ll_best - self.t_ppf(params), grad - self.t_ppf_grad(params), hess - self.t_ppf_hess(params)
+
+    def tilt_fun(self, x):
+        return -1.*x
```
pelssers, line 415: Here we should also add the direction so that the minimizer can do both upper and lower limits.
`return -self.direction * x`

flamedisx/inference.py
```diff
@@ -409,6 +469,7 @@ class IntervalObjective(Objective):
     # Add constant offset to objective, so objective is not 0 at the minimum
     # and relative tolerances mean something.
     _offset = 1
+    _callbackbag = []
```
pelssers, line 472: unused

flamedisx/likelihood.py
```diff
@@ -7,6 +7,11 @@
 import tensorflow as tf
 import typing as ty
 
+#### PL
+import pdb as pdb
+import scipy.optimize as spo
+from scipy.optimize import NonlinearConstraint
+#### end PL
```
pelssers, line 14: Unused, I guess `likelihood.py` can be left as it was. At least until we remove the tilt based interval classes at which point we can clean up the interface in `Likelihood`

flamedisx/inference.py
```diff
@@ -391,12 +396,69 @@ def parse_result(self, result: Minuit):
...
+        m2ll, grad, hess = Objective._inner_fun_and_grad(self, params)
+        return m2ll - self.m2ll_best - self.t_ppf(params), grad - self.t_ppf_grad(params), hess - self.t_ppf_hess(params)
+
+    def tilt_fun(self, x):
+        return -1.*x
+
+    def hess_constraint(self, x, v):
+        # TODO: Must return hessian matrix of dot(fun, v), is that what this
+        # does?
+        return np.dot(self(x).hess, v)
```
pelssers, line 422: Currently tests are failing here:

```python
return np.dot(self(x).hess, v)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (array([[1.87668812e+05, 1.82008243e-05],
       [1.82017684e-05, 2.03541267e-10]]), array([0.]))
kwargs = {}
relevant_args = (array([[1.87668812e+05, 1.82008243e-05],
       [1.82017684e-05, 2.03541267e-10]]), array([0.]), None)

>   ???
E   ValueError: shapes (2,2) and (1,) not aligned: 2 (dim 1) != 1 (dim 0)
```

pelssers:
I've added our `one_parameter_interval` test for `optimizer='nlin'` as well (at the moment it's duplicated so we test both the original and new version but might want to make `optimzer='nlin'` default later). I had to add `require_complete_guess = False` to make it work (and we do have incomplete guess support).

flamedisx/inference.py
plt109: Thinking of adding a callback feature to record the orbit of the optimizer. `_callbackbag` is used to store these intermediate values of the optimizer but the callback function is currently broken.

flamedisx/inference.py
```diff
@@ -391,12 +396,73 @@ def parse_result(self, result: Minuit):
...
+
+    def _inner_fun_and_grad(self, params):
+        # lol who exactly is this Objective._inner_fun_and_grad(self, params)?
+        m2ll, grad, hess = Objective._inner_fun_and_grad(self, params)
+        return m2ll - self.m2ll_best - self.t_ppf(params), grad - self.t_ppf_grad(params), hess - self.t_ppf_hess(params)
+
+    def tilt_fun(self, x):
+        # so that we get a scalar back regardless of what nuisance parameters we
+        # might have
+        return -self.direction*x[0]
```
pelssers, line 419: The current value of `self.target_parameter` may not always be the first element of `x`. So we probably need something like `self._array_to_dict(x)[self.target_parameter]` instead of `x[0]`

JelleAalbers: Pueh solved this in 406d289d169bd2736f46b3618ce04fe93ef38a27

flamedisx/inference.py
JelleAalbers: Looks like this was resolved (0677d4722320fc89d3445322cbd060c77ed5f818 ?) since the tests are passing

JelleAalbers:
Thanks for this Pueh. I removed the debug statements and accidental whitespace changes in other parts of flamedisx. Even if this doesn't work perfectly, I'd suggest we merge this as an 'experimental feature' so it does not get lost when we continue development.

@plt109 Could you fix the `NonlinearIntervalObjective.__init__`'s indentation to match [PEP8](https://www.python.org/dev/peps/pep-0008/) / the rest of flamedisx? I can do it but then the git blame would point the entire code to me :-) (There is also an unused tilt_overshoot argument with a hardcoded default.)

PR #72, opened 2020-05-20
Dimension scaling inside optimizers

pelssers:
This PR addresses part of issue https://github.com/FlamTeam/flamedisx-notebooks/issues/2

This re-implements the normalization of parameters used by the different optimizers. We had implemented this before but removed it in #40 following a discussion in #36 without really understanding the root cause and assuming that the optimizers work well also with unnormalized/physical parameters. Using SciPy's `trust-constr` method (the current default, because it takes bounds and the hessian) we see in the [Tutorial notebook](https://github.com/FlamTeam/flamedisx-notebooks/blob/master/Tutorial.ipynb) that it has a problem when parameters have values which differ by orders of magnitudes (elife and rate_multiplier). See issue https://github.com/FlamTeam/flamedisx-notebooks/issues/2
Using normalized dimensions fixes this problem.

For example, when optimizing the likelihood of the Tutorial notebook in section '2) Inference', it now correctly varies the elife parameter, where this was stuck at the guess before. The below table shows the output of `Likelihood.bestfit` with `get_history=True` for 250 events. History now includes both scaled and unscaled gradients (scroll right in the table) so we can check the gradient values.

|    |   er_rate_multiplier |   elife |       y |   er_rate_multiplier_grad |   elife_grad |   er_rate_multiplier_scaled_grad |   elife_scaled_grad |
|---:|---------------------:|--------:|--------:|--------------------------:|-------------:|---------------------------------:|--------------------:|
|  0 |            0.03      |  410000 | 3474.77 |          -12508.3         | -0.0022088   |                   -375.248       |      -905.61        |
|  1 |            0.0564236 |  440836 | 3212.05 |           -5855.56        | -0.000404039 |                   -175.667       |      -165.656       |
|  2 |            0.100181  |  449330 | 3039.87 |           -2557.07        | -2.20264e-05 |                    -76.712       |        -9.03082     |
|  3 |            0.160433  |  449728 | 2941.41 |            -960.061       | -4.78694e-08 |                    -28.8018      |        -0.0196265   |
|  4 |            0.218464  |  449606 | 2908.26 |            -254.734       | -4.0518e-09  |                     -7.64202     |        -0.00166124  |
|  5 |            0.247048  |  449546 | 2904.33 |             -29.1218      |  9.61336e-10 |                     -0.873655    |         0.000394148 |
|  6 |            0.251277  |  449537 | 2904.27 |              -0.0992851   |  9.73159e-10 |                     -0.00297855  |         0.000398995 |
[...]

PR #76, opened 2020-07-10
Optimization improvements

JelleAalbers:
* Fix a bug in the guess of the scipy optimizers. After the scaling in #72, the guess is not necessarily an array of ones; depending on the signs of the defaults, a guess can also be 0 or -1. With this fix, I don't see the optimizer asking for out-of-bounds values anymore in the tutorial fits.
  * Fix a <= vs < bug that prevented the optimizer from evaluating likelihoods exactly at upper bounds; we just slapped it with NaN if it tried.
  * Use the initial parameter specification by the user as default bounds on all variables. E.g.if you specify `elife=(300e3, 500e3, 2)`, elife isn't allowed to go outside [300e3, 500e3] anymore. To turn off this behaviour, you can either pass `bounds_specified=False` to `Likelihood.__init__`, or set the `bounds` dictionary for any particular optimization manually.
  * Give more descriptive warnings when an optimizer gets an invalid value (NaN or infinity in the objective, gradient, or Hessian)

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -253,16 +253,17 @@ class Objective:
             self._history.append(dict(params=params, y=y,
                                       scaled_grad=grad, grad=result[1]))
 
+        x_desc = dict(zip(self.arg_names, x))
         if np.isnan(y):
-            warnings.warn(f"Objective at {x} is Nan!",
+            warnings.warn(f"Objective at {x_desc} is Nan!",
                           OptimizerWarning)
             result = self.nan_result()
         elif np.any(np.isnan(grad)):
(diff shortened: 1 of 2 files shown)


JelleAalbers:
@CristianAntochi You reran our benchmarks with this branch, and did not see a big difference, right?

I would still say these changes are for the better: the two bugfixes for sure, and using bounds by default would hopefully help for parameters like g2. Here our scaling by division means the only sensible values are very close to 1, but witjhout bounds, the optimizer does not know that. @pelssers, would you agree?

CristianAntochi:
I did rerun the benchmarks with this branch and no jump was observed as well as no statistically big difference.

PR #77, opened 2020-07-14
X1T ER fits with reconstruction bias mean to simulated events

plt109:
Added functions,`recon_bias_s1` and `recon_bias_s2` along with associated helper functions, to account for pax reconstruction bias mean during simulation of events in `_simulate_response`. 

Changed value of `mean_nexni` in `p_electron` in `SR1ERSource` in `x1t_sr1.py` from 0.135 to 0.15 in accordance to value in table II of analysis paper II (1902.11297v2).


flamedisx/er_nr_base.py
```diff
@@ -195,6 +198,16 @@ def electron_gain_mean(z, *, g2=20):
     photon_gain_std = 0.5
     double_pe_fraction = 0.219
 
+    # Dummy method for pax reconstruction bias mean.
+    # Implement it in source specific class. See x1t_sr1.py till exempel
+    def recon_bias_s1(self, sig, bias_pivot_pt=1.):
```
JelleAalbers, line 203: sig -> s1? `bias_pivot_pt` is not used; users can add extra parameters in their overrides of this function

flamedisx/er_nr_base.py
```diff
@@ -659,10 +674,12 @@ class ERSource(LXeSource):
     def _single_spectrum(self):
         """Return (energies in keV, rate at these energies),
         """
+        fac = 7.
```
JelleAalbers, line 677: ???

flamedisx/x1t_sr1.py
```diff
@@ -2,11 +2,15 @@
 """
 import numpy as np
 import tensorflow as tf
+import tensorflow_probability as tfp
 
 from multihist import Hist1d
 import wimprates
 
 import flamedisx as fd
+import os
```
JelleAalbers, line 11: Have a look at the import order: https://www.python.org/dev/peps/pep-0008/#imports

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
-
 s1_map, s2_map = [
     fd.InterpolatingMap(fd.get_resource(fd.pax_file(x)))
     for x in ('XENON1T_s1_xyz_ly_kr83m-SR1_pax-664_fdc-adcorrtpf.json',
               'XENON1T_s2_xy_ly_SR1_v2.2.json')]
 
+##
+# Parameters
+##
+def_g1 = 0.142
```
JelleAalbers, line 30: Default constants are ideally named in all caps, and written out fully, e.g. DEFAULT_G1, DEFAULT_AREA_FRACTION_TOP, etc. 

It is likely a bad idea, but you could consider grouping all of these in a single [immutabledict](https://pypi.org/project/immutabledict/). An ordinary dictionary won't work, then someone will think they can set the entries and then wonder why the function defaults don't change (A: because they are evaluated when the function is defined).

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+def_g2 = 11.4 # don't divide by weird shits. will explode downstream.
```
JelleAalbers, line 31: Comment is and not very enlightening (and not very polite)

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+path_recon_bias_mean_s2 = ['/home/peaelle42/software/bbf/bbf/data/ReconstructionS2BiasMeanLowers_SR1_v2.json',
+         '/home/peaelle42/software/bbf/bbf/data/ReconstructionS2BiasMeanUppers_SR1_v2.json']
+
+path_dummy_zeros_s1 = ['dummy_zeros_s1.json', 'dummy_zeros_s1.json']
+path_dummy_zeros_s1 = [dummy_base+x for x in path_dummy_zeros_s1]
+path_dummy_zeros_s2 = ['dummy_zeros_s2.json', 'dummy_zeros_s2.json']
+path_dummy_zeros_s2 = [dummy_base+x for x in path_dummy_zeros_s2]
+
+def read_bias_tf(path_bag):
+    # Achtung:
```
JelleAalbers, line 60: Add docstring? Maybe `paths` instead of `path_bag`?

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+def cal_bias_tf(sig, fmap, support_def, pivot_pt):
+    tmp = tf.convert_to_tensor(sig, dtype=fd.float_type())
+    bias_low = tfp.math.interp_regular_1d_grid(x=tmp, 
+            x_ref_min=support_def[0], x_ref_max=support_def[1], y_ref=fmap[0],
+            fill_value='constant_extension')
+    bias_high = tfp.math.interp_regular_1d_grid(x=tmp, 
+            x_ref_min=support_def[0], x_ref_max=support_def[1], y_ref=fmap[1],
+            fill_value='constant_extension')
+
+    tmp = tf.math.subtract(bias_high, bias_low)
```
JelleAalbers, line 82: `bias_high - bias_low` should be equivalent but easier to read; similarly below

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+# Flamedisx sources
+##
 class SR1Source:
-    drift_velocity = 1.335 * 1e-4   # cm/ns
+    drift_velocity = def_drift_vel
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        try:
```
JelleAalbers, line 114: How about just loading the dummy maps in this PR, and leave the bbf map loading to after we've merged with master? Then you don't need this duplicated code. 

A catch-all try-except without re-raise is dangerous in general (e.g. it also catches SyntaxError; maybe the code just never runs), though I understand this code is just a placeholder and the print statements tell you which path you took. You don't want 10+ print statements in a production code that loads several sources though.

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+            # Loading combined cut acceptances
+            self.cut_accept_map_s1, self.cut_accept_support_s1 = read_bias_tf(path_dummy_ones_s1)
+            self.cut_accept_map_s2, self.cut_accept_support_s2 = read_bias_tf(path_dummy_ones_s2)
+            print('Dummy combined cut acceptances maps loaded')
+            # Loading reconstruction bias map
+            self.recon_map_s1_tf, self.support_def_s1 = read_bias_tf(path_dummy_zeros_s1)
+            self.recon_map_s2_tf, self.support_def_s2 = read_bias_tf(path_dummy_zeros_s2)
+            print('Dummy reconstruction bias mean maps loaded')
+
+    def recon_bias_s1(self, sig, bias_pivot_pt=def_recon_pivot):
```
JelleAalbers, line 133: Maybe bias_offset or bias_mean_shift instead of bias_pivot_pt?

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
-    def photon_detection_eff(s1_relative_ly, g1 =0.142):
-        #g1 = 0.142 from paper
-        mean_eff= g1 / (1. + 0.219)
+    def photon_detection_eff(s1_relative_ly, g1 =def_g1):
+        mean_eff= g1 / (1. + def_p_dpe)
         return mean_eff * s1_relative_ly
 
+    def s1_acceptance(self, s1, photon_detection_eff, photon_gain_mean, mean_eff=def_g1 / (1 + def_p_dpe), 
+            cs1_min=3, cs1_max=70):
+        print('s1_acceptance: cs1 min = %i, cs1 max = %f' % (cs1_min, cs1_max))
```
JelleAalbers, line 185: Leftover print statement

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
+        cs1 = mean_eff * s1 / (photon_detection_eff * photon_gain_mean)
+        mask = tf.where((cs1 > cs1_min) & (cs1 < cs1_max),
+                        tf.ones_like(s1, dtype=fd.float_type()),
+                        tf.zeros_like(s1, dtype=fd.float_type()))
+
+        # multiplying by combined cut acceptance
+        cut_accept_wgt = itp_cut_accept_tf(s1, self.cut_accept_map_s1, self.cut_accept_support_s1)
+        mask_out = tf.math.multiply(mask, cut_accept_wgt)
```
JelleAalbers, line 193: Similar to earlier, just * works and reads better

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
 
+        return mask_out
+
+    def s2_acceptance(self, s2, electron_detection_eff, electron_gain_mean,
+        cs2b_min=50.1, cs2b_max=7940):
+        print('s2_acceptance: cs2b min = %i, cs2b max = %f' % (cs2b_min, cs2b_max))
```
JelleAalbers, line 199: Leftover print statement

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
     @staticmethod
-    def photon_detection_eff(s1_relative_ly, g1 =0.142):
-        #g1 = 0.142 from paper
-        mean_eff= g1 / (1. + 0.219)
+    def photon_detection_eff(s1_relative_ly, g1 =def_g1):
+        mean_eff= g1 / (1. + def_p_dpe)
         return mean_eff * s1_relative_ly
 
+    def s1_acceptance(self, s1, photon_detection_eff, photon_gain_mean, mean_eff=def_g1 / (1 + def_p_dpe), 
+            cs1_min=3, cs1_max=70):
```
JelleAalbers, line 184: Could consider defining these as DEFAULT_CS1_MIN along with all the other defaults. Or leave all defaults defined with the function itself (which is not so bad) -- but mixing the two can be confusing

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
+                        tf.zeros_like(s1, dtype=fd.float_type()))
+
+        # multiplying by combined cut acceptance
+        cut_accept_wgt = itp_cut_accept_tf(s1, self.cut_accept_map_s1, self.cut_accept_support_s1)
+        mask_out = tf.math.multiply(mask, cut_accept_wgt)
 
+        return mask_out
+
+    def s2_acceptance(self, s2, electron_detection_eff, electron_gain_mean,
+        cs2b_min=50.1, cs2b_max=7940):
```
JelleAalbers, line 198: Similar to above. s1_acceptance and s2_acceptance look very similar, could we factor out the common code in a function?

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
+
+        return mask_out
 
 # ER Source for SR1
 @export
 class SR1ERSource(SR1Source,fd.ERSource):
 
     @staticmethod
-    def p_electron(nq, W=13.8e-3, mean_nexni=0.135,  q0=1.13, q1=0.47,
+    def p_electron(nq, W=13.8e-3, mean_nexni=0.15,  q0=1.13, q1=0.47,
```
JelleAalbers, line 216: Could you comment in the PR why you changed the values? (I think you did it for a good reason, but best to have a record of why somewhere)

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+
+def_recon_pivot = 0.49
 
 ##
-# Flamedisx sources
+# Loading Pax reconstruction bias
+##
+dummy_base = os.path.dirname(os.path.realpath(__file__)) + '/dummy_maps/'
+
+path_recon_bias_mean_s1 = ['/home/peaelle42/software/bbf/bbf/data/ReconstructionS1BiasMeanLowers_SR1_v2.json',
```
JelleAalbers, line 49: See below -- maybe just leave out the bbf map loading until we merge. Your home dir won't exist for any of us

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+def read_bias_tf(path_bag):
+    # Achtung: 
+    # Fundamentally assumes upper and lower bounds have exactly the same support definition
+    data_bag = []
+    yy_ref_bag = []
+    for loc_path in path_bag:
+        with open(loc_path) as json_file:
+            tmp = json.load(json_file)
+            yy_ref_bag.append(tf.convert_to_tensor(tmp['map'], dtype=fd.float_type()))
+            data_bag.append(tmp)
```
JelleAalbers, line 68: These two lines could be deindented so the json can be closed immediately after loading tmp. (and, more importantly, it's clear you're done with the file after the first line)

flamedisx/x1t_sr1.py
```diff
@@ -15,21 +19,126 @@
...
+    support_def = tmp['coordinate_system'][0][1] 
+    
+    return yy_ref_bag, support_def
+
+def cal_bias_tf(sig, fmap, support_def, pivot_pt):
```
JelleAalbers, line 73: Please add a docstring, and ideally use more descriptive argument names. We certainly haven't done this everywhere in flamedisx, but it is good practice :-)

flamedisx/x1t_sr1.py
```diff
@@ -53,37 +162,61 @@ def add_extra_columns(self, d):
...
+    def s1_acceptance(self, s1, photon_detection_eff, photon_gain_mean, mean_eff=def_g1 / (1 + def_p_dpe), 
+            cs1_min=3, cs1_max=70):
+        print('s1_acceptance: cs1 min = %i, cs1 max = %f' % (cs1_min, cs1_max))
+        cs1 = mean_eff * s1 / (photon_detection_eff * photon_gain_mean)
+        mask = tf.where((cs1 > cs1_min) & (cs1 < cs1_max),
+                        tf.ones_like(s1, dtype=fd.float_type()),
+                        tf.zeros_like(s1, dtype=fd.float_type()))
+
+        # multiplying by combined cut acceptance
+        cut_accept_wgt = itp_cut_accept_tf(s1, self.cut_accept_map_s1, self.cut_accept_support_s1)
```
JelleAalbers, line 192: wgt -> weight is just a few more characters. Though maybe you don't need more than one variable here. Just do 
```
acceptance = tf.where(...)
acceptance *= itp_cut...
return acceptance
```
Or even take the last two lines into one line (return acceptance * itp_cut...), but this is a matter of taste. I have been known to write too long lines (bad practice from Mathematica days...).

flamedisx/er_nr_base.py
```diff
@@ -723,6 +740,7 @@ def _simulate_nq(self, energies):
         return np.floor(energies / work).astype(np.int)
 
 
+
```
JelleAalbers, line 743: Spurious newline

flamedisx/er_nr_base.py
```diff
@@ -305,7 +318,7 @@ def _differential_rate(self, data_tensor, ptensor):
 
         p_ph = self.detection_p('photon', data_tensor, ptensor)
         p_el = self.detection_p('electron', data_tensor, ptensor)
-        d_ph = self.detector_response('photon', data_tensor, ptensor)
+        d_ph = self.detector_response('photon', data_tensor, ptensor) # p(s1|ph_detected)
```
JelleAalbers, line 321: Thanks for adding the comment. You could consider putting it on a separate line (so the line itself stays within 80char), and /or adding similar comments for the other three similar lines?

flamedisx/er_nr_base.py
```diff
@@ -26,9 +26,12 @@
     'p_electron_fluctuation',
     'electron_acceptance',
     'photon_acceptance',
-    'penning_quenching_eff'
+    'penning_quenching_eff',
+    'recon_bias_s1',
```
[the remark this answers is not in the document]
plt109: edited.

flamedisx/er_nr_base.py
```diff
@@ -26,9 +26,12 @@
...
+    'recon_bias_s2'
 ]
 
+#
```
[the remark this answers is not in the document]
plt109: removed.

flamedisx/er_nr_base.py
```diff
@@ -195,6 +198,16 @@ def electron_gain_mean(z, *, g2=20):
     photon_gain_std = 0.5
     double_pe_fraction = 0.219
 
+    # Dummy method for pax reconstruction bias mean.
+    # Implement it in source specific class. See x1t_sr1.py till exempel
```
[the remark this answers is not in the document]
plt109: Added doc string.

flamedisx/x1t_sr1.py
plt109: Should we print out some of the higher level settings? Eg. cuts applied in cs1, cs2, etc.

flamedisx/x1t_sr1.py
plt109: g2 bottom as defined in 1902.11297v2. instead of g2 total.

flamedisx/x1t_sr1.py
plt109: added docstring. kept `paths` as `path_bag` for ease of reading to tell that it is an array of paths; one does not need to look out for 's' at the end of the variable.

flamedisx/x1t_sr1.py
plt109: Kept it for explicit usage of tf.math.subtract instead of implicit overloading of `-` operator.

flamedisx/x1t_sr1.py
plt109: Changed value of `mean_nexni` in `p_electron` in `SR1ERSource` in `x1t_sr1.py` from 0.135 to 0.15 in accordance to value in table II of analysis paper II (1902.11297v2).

Updated PR to include this.

flamedisx/er_nr_base.py
plt109: Removed spurious new line. So 2 blank lines between 2 consecutive classes in the same .py?

flamedisx/x1t_sr1.py
plt109: Shifted to top.

flamedisx/x1t_sr1.py
plt109: There is a lower bound, `bias_low`, and upper bound, `bias_high`, to the reconstruction bias mean as there is an uncertainty to this bias mean due to its dependence on the waveform. The final reconstruction bias mean is calculated using a single parameter `bias_pivot_pt` to linearly tune where between this lower and upper bound, the reconstruction bias mean is.

`bias_out` = `bias_pivot_pt`*(`bias_high`-`bias_low`) + `bias_low`

So in that sense, it is like the 'pivoting point'.

flamedisx/er_nr_base.py
plt109: Extends max input ER energy to 7*10 keV and keeps the number of bins constant at 1000 bins per 10 keV. Changed variable name from `fac` to `max_e`.

flamedisx/x1t_sr1.py
plt109: Parameters that are used across more than 1 function(with the exception of elife and drift velocity, since they are important and would be good to have them visible at one glance) are defined as global variables at the top. Decided to keep cs1_min, cs1_max, cs2_min, cs2_max defined within the function for now.

flamedisx/x1t_sr1.py
plt109: Yes, we should decide which variables to keep inside the 'common block'.

flamedisx/x1t_sr1.py
plt109: Edited.

flamedisx/x1t_sr1.py
plt109: Prefer explicit usage of tf.math.multiply to implicit overloading of `*` operator.

flamedisx/x1t_sr1.py
plt109: Agree, but would prefer to have a copy that works locally to make sure changes have been propagated properly by checking that the bias corrections are still applied. The local paths will disappear once the additional functions from bbf_loading is available.

flamedisx/x1t_sr1.py
plt109: Wanted to explicitly use `tf.math.multiply` for explicitness.

JelleAalbers: We discussed and changed it to `*`, for consistency with the rest of the code.

Issue #68, opened 2020-03-23
Full accounting of double-photoelectron emission


JelleAalbers:
Currently flamedisx accounts for double-PE emission approximately, by increasing the mean and width of the Gaussian S1 resolution. By the central limit theorem, this reaches any desired level of accuracy above some number of photons -- in practice it's reasonable for more than ~5. 

... but of course, DM searches use < 5 PE signals, and it wouldn't be particularly hard to add an extra binomial step in flamedisx to account for DPE emission as best we can.

JelleAalbers:
On the shorter term, we might want to check if our approximate formula is correct for non-unity mean PE gains: it's easy to miss this factor in one of the formulas since it is usually 1.

PR #78, opened 2020-07-15
Double photoelectron emission modeling

pelssers:
This implements double photoelectron emission (dpe) modeling using a binomial process instead of the previous gaussian approximation making the model more realistic for low energy S1s.
The probability of the number of dpe's is modeled as: `n_pe - n_photons ~ Binom(n_photons, p=p_dpe)` where `n_pe` is the number of photoelectrons detected, `n_photons` is the number of photons detected and `p_dpe` is de dpe fraction.
To implement this an extra tensor of shape `(n_events, n_photoelectron detected, n_photon detected)` is added to the core flamedisx model and an extra inner dimension `photoelectron_detected` is added. Bounds estimation and simulation is added as well.

The changes have been validated using the [models_check notebook](https://github.com/FlamTeam/flamedisx-notebooks/blob/master/models_check.ipynb)

This implements a solution for issue #68

### DIFF
--- a/flamedisx/er_nr_base.py
+++ b/flamedisx/er_nr_base.py
@@ -18,7 +18,7 @@ export, __all__ = fd.exporter()
 o = tf.newaxis
 
 quanta_types = 'photon', 'electron'
-signal_name = dict(photon='s1', electron='s2', photoelectron='s1')
+signal_name = dict(photon='s1', electron='s2')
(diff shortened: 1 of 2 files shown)


pelssers:
Thanks for reviewing and further improvements! Looks good to me, merging.

PR #80, opened 2020-07-19
Access BBF data, reorganize XENON-specific utilities

JelleAalbers:
This is an update to the interface to access data from XENON repositories.
  * Everything XENON-specific is now in the `xenon` subfolder/package. It is still exported to the flamedisx package, so you can keep doing `fd.SR1ERSource` etc.
  * Updated the resource caching and correction maps from straxen, and moved each to a separate file.
  * Added an interface with BBF, see below.

This adds the `fd.get_bbf_file` function, which fetches data from BBF. Use as, for example, `fd.get_bbf_file('ACBackgroundSR0.json')`, which fetches https://github.com/XENON1T/bbf/blob/master/bbf/data/ACBackgroundSR0.json and parses it as dictionary.

BBF is a private repository, so loading these files requires authentication. There are three ways this can work:
  * You can do nothing special and just run flamedisx. If you use a source which needs a bbf file, it will instruct you to create a Github authentication token, with which flamedisx will clone bbf, then check out the file.
  * Alternatively, call `fd.ensure_bbf('........................................')` in your notebook, with your token in place of the dots. This will clone bbf immedatiately, so future `get_bbf` calls succeed without prompting.
  * Alternatively, clone bbf yourself somewhere, then set `fd.XENON.data.BBF_PATH` to the path of bbf's root folder. For example, I do `fd.XENON.data.BBF_PATH = '/home/aalbers/software/bbf'`.

There is definitely room for improvement here. Some considerations:
  * We're cloning the entire 800 MB bbf repo, since it is not straightforward to get a single file in a private repo with github. If you need one BBF file, you will likely need many, so it is not so bad. Cloning with `--depth 1` only saves ~100 MB; most of BBF's size is in a few static data files.
  * We could have used a username/password in the git clone URL, instead of the OAuth token stuff. However, if git clone encounters any problem, it prints the URL to the screen, including your plaintext password...
  * We could consider requiring straxen in flamedisx, so the `itp_map.py` and `resource.py` files can be removed. However, straxen depends on strax, which has some dependencies that don't reliably distribute working pip binaries (or at least, for some reason straxen's Travis CI spends 10 minutes building C++ code...). Maybe these files could be moved to some common lightweight XENON repo like https://github.com/XENONnT/utilix.


flamedisx/xenon/itp_map.py
```diff
@@ -0,0 +1,166 @@
...
+
+import numpy as np
+from scipy.spatial import cKDTree
+
+import flamedisx as fd
+export, __all__ = fd.exporter()
+
+
+@export
+class InterpolateAndExtrapolate:
```
plt109, line 18: Reminiscent of `InterpolateAndExtrapolate` and `InterpolatingMap` in xenon_corrections.py but with minor changes to `InterpolatingMap`.

flamedisx/utils.py
```diff
@@ -203,3 +206,17 @@ def values_to_constants(kwargs):
 def wilks_crit(confidence_level):
     """Return critical value from Wilks' theorem for upper limits"""
     return stats.norm.ppf(confidence_level) ** 2
+
+
+@export
+def run_command(command):
+    """Run command and show its output in STDOUT"""
+    # Is there no easier way??
```
plt109, line 214: For cleaner syntax and simple operations, it is possible to use `os.system(command)` instead as suggested in https://stackoverflow.com/questions/4813238/difference-between-subprocess-popen-and-os-system. However, official documentation for `os` module suggests using `subprocess` module https://docs.python.org/3/library/os.html. So, keeping software longevity in mind, it is probably better to use `subprocess`.

JelleAalbers: Thanks Pueh; I tried os.system, but it didn't show me the output of the command while it was running. Since the cloning of BBF takes some time, I thought it was nicer to show the output as it happened, like ! from ipython. But this indeed comes at the cost of some extra complication.

flamedisx/xenon/resource.py
```diff
@@ -0,0 +1,160 @@
...
+
+# In-memory resource cache
+_resource_cache = dict()
+
+# Formats for which the original file is text, not binary
+_text_formats = ['text', 'csv', 'json']
+
+
+@export
+def get_resource(x, fmt=None):
```
plt109, line 29: Reminiscent of `get_resource` in xenon_corrections.py

flamedisx/xenon/itp_map.py
JelleAalbers: Indeed, this is the same code with minor changes (backported from straxen), moved to a different file

JelleAalbers:
Thanks for the comments Pueh! Hope this will make your life easier implementing the various XENON files.

PR #82, opened 2020-08-03
Pax recon eff

plt109:
Added `cal_rec_efficiency_tf` in x1t_sr1.py which is called in `photon_acceptance`.

Previously, there is only a requirement for the event to have at least 3 photons detected and a flat unity acceptance thereafter. 

In `cal_rec_efficiency_tf`, the PAX reconstruction efficiency is interpolated from a weighted combination of median and lower/upper bounds similar to the logic in BBF, see https://github.com/XENON1T/bbf/blob/master/bbf/gpu_modules/SignalSimCuda.py#L483. This reconstruction efficiency is then applied to the events via `photon_acceptance`.

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -44,26 +44,34 @@ DEFAULT_S2_RECONSTRUCTION_BIAS_PIVOT = 0.49198507921078005
 DEFAULT_S1_RECONSTRUCTION_EFFICIENCY_PIVOT = -0.31816407029454036 
 
 ##
-# Loading Pax reconstruction efficiencies (do not reorder)
+# Combined cuts acceptances
 ##
-path_reconstruction_efficiencies_s1 = ['RecEfficiencyLowers_SR1_70phd_v1.json',
-       'RecEfficiencyMedians_SR1_70phd_v1.json',
-       'RecEfficiencyUppers_SR1_70phd_v1.json']
+path_cut_accept_s1 = ['S1AcceptanceSR1_v7_Median.json']
+path_cut_accept_s2 = ['S2AcceptanceSR1_v7_Median.json']
 
 ##
-# Loading Pax reconstruction bias
-## bins=nbins,
+# Pax reconstruction bias (do not reorder: Lowers, Uppers)
+## 
 path_reconstruction_bias_mean_s1 = ['ReconstructionS1BiasMeanLowers_SR1_v2.json',
-        'ReconstructionS1BiasMeanUppers_SR1_v2.json']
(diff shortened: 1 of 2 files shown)


flamedisx/xenon/x1t_sr1.py
```diff
@@ -38,24 +38,40 @@
...
 path_reconstruction_bias_mean_s2 = ['ReconstructionS2BiasMeanLowers_SR1_v2.json',
-        'ReconstructionS2BiasMeanUppers_SR1_v2.json']
+                                    'ReconstructionS2BiasMeanUppers_SR1_v2.json']
+
+##
+# Pax reconstruction efficiencies (do not reorder: Lowers, Medians, Uppers)
+##
+path_reconstruction_efficiencies_s1 = ['RecEfficiencyLowers_SR1_70phd_v1.json',
+                                       'RecEfficiencyMedians_SR1_70phd_v1.json',
+                                       'RecEfficiencyUppers_SR1_70phd_v1.json']
```
pelssers, line 65: If the order of the filenames if important I would consider using a dict instead of a list. That way you'll always get the right file for a given key instead of relying on the index in the list. With a dict you can also combine all the different global `path_*` variables for cut acceptances, biases and efficiencies and have them neatly in one place.
It does mean you have to change 'read_maps_tf' to work with the dict.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -96,18 +125,32 @@ def cal_bias_tf(sig, fmap, domain_def, pivot_pt):
...
+    :param fmap: map returned by read_maps_tf
+    :param domain_def: domain returned by read_maps_tf
+    :param pivot_pt: Pivot point value (scalar)
+    :return: Tensor of bias values (same shape as sig)
+    """
+    tmp = tf.convert_to_tensor(sig, dtype=fd.float_type())
 
+    bias_median = tfp.math.interp_regular_1d_grid(x=tmp,
+            x_ref_min=domain_def[0], x_ref_max=domain_def[1], y_ref=fmap[1],
+            fill_value='constant_extension')
```
pelssers, line 140: I'd split this off into a separate function since these lines are duplicated several times (lines 98, 113, 116, 138, 143, 148). Maybe something like:
```python
def tf_interpolated(x, y, domain):
    x_tf = tf.convert_to_tensor(x, dtype=fd.float_type())  # Not sure if this line is needed in every case though
    return tfp.math.interp_regular_1d_grid(x=x_tf,
            x_ref_min=domain[0], x_ref_max=domain[1], y_ref=y,
            fill_value='constant_extension')
```

JelleAalbers:
This is over a year old now, so I think we can safely close this. Feel free to reopen or make a new PR!

Issue #37, opened 2019-10-20
Numerical error in beta_binom_pmf for high means and low fluctuations


JelleAalbers:
Flamedisx currently overestimates the differential rate at low-energy NR events under the ER model. From adapting the models_check notebook on my laptop, fixing the position at the center of the TPC, and using a 30 GeV WIMPsource, I get:

![nrdata_ermodel](https://user-images.githubusercontent.com/4354311/67161479-3f3c4a00-f35b-11e9-985d-246ab4ad7bc1.png)

This has to be the reason why, in our discrimination efficiency check, flamedisx is under some conditions outperformed by a classic (cS1, cS2) likelihood. 

Since flamedisx *over*estimates the rate, the problem can't be in the bounds calculation, so increasing max_sigma does not help.

All the other combinations (ER data + ER model, ER data + NR model, NR data + NR model, ) give consistent results:

![erdata_ermodel](https://user-images.githubusercontent.com/4354311/67161486-47948500-f35b-11e9-99fc-ec0bd63eb7a0.png)
![erdata_nrmodel](https://user-images.githubusercontent.com/4354311/67161487-47948500-f35b-11e9-82a7-69d3e4bf9718.png)
![nrdata_nrmodel](https://user-images.githubusercontent.com/4354311/67161488-49f6df00-f35b-11e9-85ad-57673e53b54d.png)

JelleAalbers:
The problem is due to numerical errors / instabilities in fd.beta_binom_pmf. When the fluctuation amplitude becomes small, especially when the mean probability goes to 1, we get an incorrect PMF. Here is the case for mean = 97% and fluctuation = 1%. If you make the values more extreme you can get more crazy distributions.

```python
import matplotlib.pyplot as plt
import numpy as np
from multihist import Hist1d

from scipy import stats
import flamedisx as fd

# These might be introduced by the next PR
fd.MIN_FLUCTUATION_P = 0.
fd.MAX_MEAN_P = 1.

mean, sigma, n  = 0.97, 0.01, 20
n_trials = int(1e6)

a, b = fd.beta_params(mean, sigma)
p_means = stats.beta(a, b).rvs(n_trials)
ks = stats.binom(n=n, p=p_means).rvs(n_trials)
bins = np.arange(-2, 1.5 * n + 2) - 0.5
h = Hist1d(ks, bins) / n_trials
h.plot()

plt.scatter(h.bin_centers, 
[...]

PR #83, opened 2020-08-03
Fix beta-binomial bug and bimodality

JelleAalbers:
This fixes two issues in flamedisx's beta bimomial:
  1. We incorrectly computed the beta distribution's alpha and beta given a mean and sigma;
  2. We allowed for bimodal beta distributions, for sufficiently high fluctuations.

I will call alpha = a, beta = b, mean = m, variance = sigma**2 = s for simplicity.

## Alpha beta computation bug
We assumed v = 1/(8 b + 4), but this only holds exactly when a = b. The general expression is v = a b / ((a + b)^2 (a + b + 1)). We can still solve (a, b) given (m, s), though it is [slightly hairier](https://www.wolframalpha.com/input/?i=solve+v+%3D%3D+a+b+%2F+%28%28a+%2B+b%29%5E2+%28a+%2B+b+%2B+1%29%29+and+m+%3D%3D+a%2F%28a%2Bb%29+for+a+and+b).

We did not notice this earlier because  
 - Recombination fluctuations are an extremely small effect at low energies, compared to statistical fluctuations. Given that our current formula is pretty accurate most of the time, the discrepancy with any real data we looked at is very small.
 - Since we use the same (m, s) -> (a, b) conversion in the simulator and the direct computation, an effect like this would never be picked up by the (more sensitive) direct vs MC differential rate computation.

## Bimodality
The beta distribution becomes undefined for a < 0 or b < 0. The solution linked above shows this happens outside a radius 1/2 semicircle around (0.5, 0) in the (m, s) plane.

A bigger constraint is that we want the beta distribution for physical probabilities to be unimodal. However, if max(a, b) < 1, the PDF becomes bimodal -- it will have two local maxima, at 0 and 1 (not necessarily equally large). An extreme case of this is the a = b = 0.5 curve shown in [red on Wikipedia's Beta distribution article](https://en.wikipedia.org/wiki/Beta_distribution).

The bimodality happens on the outer edge of the valid semicircle:

beta_bimodality

The green curves can also be derived from the m and s expressions; see the code for the result. After this, flamedisx will clip s to lie under the green curve.

## Minimum mean
Since we now have a maximum on s, besides the minimum we already had (see #37), extreme values of m no longer have any legal s we can pick for them. We already capped the maximum m in #37. I now capped the minimum m at 0.011 (1.1%), since for m = 0.01 the lowest allowed s = 0.005 is just in the bimodal region.

For comparison, reported values of the binomial fluctuation s are around 4-6% (see XENON and LUX papers). In XENON's model, s goes to 0 below 2 keV (see analysis paper II), but this was likely done to avoid fluctuations outside [0, 1] for the Gaussian model. (And at 2 keV, statistical fluctuations are so large that binomial overdispersion would seem very difficult to detect.)

The unimodality constraint already excludes the dangerous region near the edge of the range of validity, so I hope we do not need an extra 'numerical safe zone' around it. (If there is be a super  small turn-up deep in the tail in the beta distribution, that should be OK.)

### DIFF
--- a/flamedisx/er_nr_base.py
+++ b/flamedisx/er_nr_base.py
@@ -383,7 +383,6 @@ class LXeSource(fd.Source):
             pel_fluct = self.gimme('p_electron_fluctuation', bonus_arg=_nq_1d,
                                    data_tensor=data_tensor, ptensor=ptensor)
             pel_fluct = fd.lookup_axis1(pel_fluct, _nq_ind)
-            pel_fluct = tf.clip_by_value(pel_fluct, fd.MIN_FLUCTUATION_P, 1.)
             return rate_nq * fd.beta_binom_pmf(
                 nph,
                 n=nq,
@@ -652,9 +651,6 @@ class LXeSource(fd.Source):
 
         if self.do_pel_fluct:
             d['p_el_fluct'] = gimme('p_electron_fluctuation', d['nq'].values)
-            d['p_el_fluct'] = np.clip(d['p_el_fluct'].values,
-                                      fd.MIN_FLUCTUATION_P,
-                                      1.)
             d['p_el_actual'] = 1. - stats.beta.rvs(
                 *fd.beta_params(1. - d['p_el_mean'], d['p_el_fluct']))
         else:
@@ -745,9 +741,7 @@ class ERSource(LXeSource):
     def p_electron_fluctuation(nq):
         # From SR0, BBF model, right?
         # q3 = 1.7 keV ~= 123 quanta
-        return tf.clip_by_value(0.041 * (1. - tf.exp(-nq / 123.)),
-                                fd.MIN_FLUCTUATION_P,
-                                1.)
+        return 0.041 * (1. - tf.exp(-nq / 123.))
 
     @staticmethod
     def penning_quenching_eff(nph):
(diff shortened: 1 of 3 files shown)


pelssers:
Good that you caught this bug. I was thinking if we can remove the beta binomial pmf altogether using the new 'block' implementation of the model you are developing in #81 . What if we introduced a separate 'beta' and 'binomial' block instead of doing the computation in one block, the beta binomial has also always been the bottleneck in performance so introducing a new block (for ER sources only) might not be so bad for performance and possible rid us of the numerical issues.

JelleAalbers:
That's an interesting idea; having it as an option would also allow people to change the Beta into a Gaussian more easily. I'm not sure we can do this without using 4d tensors, but if so, it would definitely be worth trying. We'd still need the fixes in this issue, as long as we use the beta somewhere.

pelssers:
We shouldn't forget to add the new bounds on mean and fluctuation to the simulator as well (just came across this while reviewing the blocks branch)

JelleAalbers:
Indeed, they should be consistent. I made the modification in beta_params, which is also called in simulate: https://github.com/FlamTeam/flamedisx/blob/master/flamedisx/er_nr_base.py#L659. But you're right that there is another clip in simulate before that, which shouldn't be necessary anymore.

pelssers:
Ok, I've removed any remaining clips on `MIN_FLUCTUATION_P` outside `beta_params`. Fine with merging if tests pass. We should take care this implementation is the same in the new blocks branch which removes `er_nr_base.py`

PR #89, opened 2020-08-17
Add cS1,cS2 computation to SR1Source

### LINKED ISSUE #88: SR1 sources not simulating
The SR1 source simulators are not working at the moment. Probably since #81 was merged which added the `cS1`, `cS2` computation. Since these sources require private XENON1T files to run we do not test them and did not catch this on Travis. 

Reproduce with:
```python
import flamedisx as fd

s = fd.SR1ERSource()
s.simulate(10)
```
will result in ```KeyError: 'cs1'```

pelssers:
This fixes #88 
It adds the cS1,cS2 computation to `add_extra_columns` of  `SR1Source`, in the same way as is done in `SR0Source`. As noted in a comment in the original implementation in `SR0Source`: 'this is a kludge'. Would it be better to add cS1 and cS2 as model functions instead?

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -172,6 +172,17 @@ class SR1Source:
                           d['y'].values,
                           d['z'].values]))
 
+        # Add cS1 and cS2 following XENON conventions.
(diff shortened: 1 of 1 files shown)


plt109 (approved):
Thanks Bart!

PR #91, opened 2020-10-16
Do not round MLE of photons detected

JelleAalbers:
The `photons_detected_mle` variable, i.e. the maximum-likelihood estimate of photons detected, was rounded to an integer. While technically correct, this causes information loss, so that other variables (e.g. `photons_produced_mle`, `e_vis_mle`) will appear coarser. It is also inconsistent with `electrons_detected_mle`, which is not rounded.

I noticed this because the final output the tutorial now looks odd:

odd

After this fix, it looks fine again:

better


plt109:
Thanks Jelle! I agree that it is better to not round `photons_detected_mle` for consistency with not rounding `electrons_detected_mle`, and also to prevent downstream artifacts in other variables like `e_vis_mle`.

PR #94, opened 2021-01-07
Minuit 2 support

JelleAalbers:
This updates flamedisx's MinuitObjective to support minuit 2.0. I kept the minuit 1 wrapper code, since the release is very new and we haven't tested much with minuit 2 yet (since I think we're usually using scipy nowadays).


plt109:
Thanks Jelle for this update!

I will just note here that this is because there is a breaking change in the iminuit package on 7 Dec 2020 to iminuit 2.0.0. See https://iminuit.readthedocs.io/en/stable/changelog.html.

PR #95, opened 2021-01-20
fix import errors

sophiaandaloro:
With current tensorflow 2.4.0, tensorflow-probability has to be 0.7.0 otherwise I get the following error:
```
from tensorflow.python.autograph.core import naming ImportError: cannot import name 'naming
```
Flamedisx imports fine when I do this. Let me know if there's a better way to solve the import issue, though

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ multihist
 wimprates>=0.3.1
 tqdm
 tensorflow>=2.2.0
-tensorflow_probability>=0.8.0
+tensorflow_probability==0.7.0
 iminuit


JelleAalbers:
Thanks for the report Sophia! Unfortunately something is wrong with our Travis configuration, so it did not test your change yet. 

I would be a bit surprised if tfp 0.7.0 works: @pelssers moved away from 0.7.0 in September 2019 (https://github.com/FlamTeam/flamedisx/pull/17) since it did not work with tensorflow 2. Tfp 0.7.0 was developed against tensorflow 1.14 (https://github.com/tensorflow/probability/releases/tag/v0.7), while flamedisx definitely needs tensorflow 2. So it's great if it works for you, but I doubt this solution will work in general.

You are not the only one with this tensorflow issue though, it seems to recur periodically and it doesn't seem like there is a clear solution:
  * https://github.com/tensorflow/tensorflow/issues/40385
  * https://github.com/tensorflow/tensorflow/issues/40584
  
On the other hand, our last trasis builds succeeded with tensorflow 2.4.0 and tensorflow-probability 0.12.1 in python 3.8 (https://travis-ci.com/github/FlamTeam/flamedisx/jobs/472192249) and 3.6 (https://travis-ci.com/github/FlamTeam/flamedisx/jobs/472192248).

So maybe it's something particular to your machine / OS / setup?

sophiaandaloro:
Yeah this issue seems to happen with different cuda packages or other weird incompatibilities. I think we probably should leave as is, as this worked on a clean environment for me on my Rice server. Tensorflow just doesn't seem to like playing nice with other packages :-)  Thanks!

PR #97, opened 2021-01-26
Copied in latest version of x1t_sr1.py with bug fix on cs2b cuts and …

plt109:
Bug fix in `s2_acceptance` to apply cut on cs2_bottom instead of cs2.

Applies pax reconstruction efficiency via method `photon_acceptance` in class `SR1Source1`.


flamedisx/xenon/x1t_sr1.py
```diff
@@ -11,13 +10,7 @@
...
 o = tf.newaxis
 
-##
-# Yield maps
-##
-s1_map, s2_map = [
-    fd.InterpolatingMap(fd.get_resource(fd.pax_file(x)))
-    for x in ('XENON1T_s1_xyz_ly_kr83m-SR1_pax-664_fdc-adcorrtpf.json',
-              'XENON1T_s2_xy_ly_SR1_v2.2.json')]
+import pdb as pdb
```
JelleAalbers, line 13: Probably left in from debugging. You can also set breakpoints with the built-in `breakpoint()`; if that's all you need you don't need to import pdb.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -68,9 +84,19 @@ def read_maps_tf(path_bag, is_bbf=False):
         yy_ref_bag.append(tf.convert_to_tensor(tmp['map'], dtype=fd.float_type()))
         data_bag.append(tmp)
     domain_def = tmp['coordinate_system'][0][1]
-
+
```
JelleAalbers, line 87: Blank line with whitespace (not sure why codefactor is sleeping on the job)

flamedisx/xenon/x1t_sr1.py
```diff
@@ -221,24 +257,26 @@ def s1_acceptance(self,
                               tf.zeros_like(s1, dtype=fd.float_type()))
 
         # multiplying by combined cut acceptance
-        acceptance *= itp_cut_accept_tf(s1,
-                                        self.cut_accept_map_s1,
-                                        self.cut_accept_domain_s1)
+        acceptance *= tf.squeeze(interpolate_tf(s1,
```
JelleAalbers, line 260: If the batch size is 1, could `tf.squeeze` cause zero-dimensional tensors to appear? It's probably safer and clearer to do a [0] on the axes you want to remove instead.

JelleAalbers (commented):
Looks good Pueh! Thanks for making some of the interpolation functions nicer, I only left some minor comments. 

I seem to have broken codefactor in https://github.com/FlamTeam/flamedisx/pull/93/commits/8f231df9edb956bef03f44e00bb3f730819eef65, so it's not even noticing things like blank lines with whitespace, or missing whitespace around some operators. But if I got a free pass on a PR, so should you ;-)

flamedisx/xenon/x1t_sr1.py
plt109: Thanks for catching. I've removed it. Nice, I'll try out breakpoint(). I just want something similar to Matlab's debugging options (breakpoints, conditional breakpoints, stepping into functions, etc).

flamedisx/xenon/x1t_sr1.py
plt109: Yes, if the batch size is 1, `tf.squeeze(interpolate_tf(s1,...))` will be a zero-dimensional tensors to appear but the `*=` makes the final `acceptance` the same dimensions as `s1`. 

I have changed it to `[0]` nevertheless for explicitness regardless.

flamedisx/xenon/x1t_sr1.py
plt109: I will let codefactor catch me in the future!

flamedisx/xenon/x1t_sr1.py
JelleAalbers: Thanks! Could you explain where the extra size-1 dimension comes from? `interpolate_tf`'s docstring says it returns the same shape as the input, but it doesn't seem to do so (s1 and s2 are 1d). In the bias functions `interpolate_tf` is used without the [0]/squeeze.

plt109: To be really honest, I have no idea either. I have a few pieces of information maybe they make sense to you but they might be totally irrelevant.

1. `interpolate_tf`, `cal_bias_tf`, `cal_bias_tf`, `cal_rec_efficiency_tf` are all written in `x1t_sr1.py` and not part of any classes. `s1_acceptance`, `s2_acceptance` are methods of the `SR1Source` class.
2. I stepped inside `interpolate_tf` and see how it behaves as it's being called by `cal_bias_tf`, `cal_bias_tf`, `cal_rec_efficiency_tf`, `s1_acceptance`, `s2_acceptance`. The shape of the input variables (`sig_tf`, `domain` and `fmap`) inside `interpolate_tf` are all the same regardless of which function called it. I also stored the output into a separate variable (instead of returning it immediately) and examined its shape. Somehow, the output variable acquired that extra dimension when `interpolate_tf` is called from `s1_acceptance` and `s2_acceptance`. 

Observation 2 is basically just restating your question as a statement without providing any extra insights on where the extra dimension came from. However, it assures that the shape of the input arguments to `interpolate_tf` are consistent regardless of where it is called from, which leads me to think that the variables(`sig_tf`, `domain` and `fmap`) are being passed around correctly. 

Superficially, I would suspect some class magic going on but much more digging (potentially inside `tfp.math.interp_regular_1d_grid` itself) will be required to really figure this thing out. This extra dimension seems to be handled by `[0]` and does not seem to present any imminent issues.

JelleAalbers: OK. I was thinking the [0] could someday select the first event's acceptance if the extra dimension disappears, and that could cause a headache.

I had a brief look: SR1Source.cut_accept_map_s1 is not a tensor, but a one-element list containing a tensor. When you pass that as `y_ref` to `interp_regular_1d_grid`, tensorflow assumes you want an output with one extra dimension (it's called `regular_1d_grid`, but accepts multidimensional tensors and silently converts lists to tensors).

Probably you either want to do [0] when you're setting the `cut_accept_map_s1` attribute (so just set it to a tensor), or if you like one-element lists, feed `self.cut_accept_map_s1[0]` to interpolate_tf instead of `self.cut_accept_map_s1`. Up to you though.

JelleAalbers (approved):

flamedisx/xenon/x1t_sr1.py
plt109: Whoa, nice catch. I did not notice the `[]` surrounding the `<tf.Tensor >` when I was stepping through the `interpolate_tf` because I was only looking at `shape` and `numpy`.

I have changed it to passing `self.cut_acceptance_map_s1[0]` into `interpolate_tf` to make it consistent with how `interpolate_tf` is called in `cal_bias_tf`, `cal_rec_efficiency_tf`.

Issue #98, opened 2021-02-02
Floating point error in the j2000 event time


jordan-palmer:
`t` can be outside of the range `self.t_start` and `self.t_end`. To fix this, I will add a clip line 

`np.clip(t, self.t_start, self.t_end)`

Might be good to add a warning If `t` is very far from this range.

JelleAalbers:
We merged a workaround in #99; #90 tracks the deeper underlying issue.

PR #99, opened 2021-02-02
Prevent CoordinateOutOfRangeException for slightly different event times

jordan-palmer:
Floating point error in the j2000 event time. To fix this, we can just clip the value of `t` to make sure it is in range (`self.t_start`,`self.t_end`). 

I submitted an issue for this too (#98 )


jordan-palmer:
I don't know the best way to add this. The line I added could just be included in the line after 

i.e 

```        result = np.stack([self.energy_hist.slicesum(np.clip(t, self.t_start, self.t_end)).histogram
                           for t in t_j2000])```

It would be nice to add a warning to this but I'm not sure what your protocol is for that.

JelleAalbers (approved):
Thanks for reporting this bug, and suggesting a fix Jordan! I made a few additions:
  * Raise a custom exception when times are 1 day out of range
  * Move the time clipping inside a function, since we need it in two places (energy_spectrum and random_truth)
  * Add a unit test for the new function
  * Tangentially related addition: allow users to fix event times with datetime objects (besides unix epoch timestamps in nanoseconds) in simulate's fix_truth.
  
Maybe this is overkill, but things get very confusing with three time conventions running around: the energy spectrum is binned in J2000 floats, event times are unix timestamp nanosecond floats, and t_start and t_stop are datetime objects.

Let me know if you agree / have any other suggestions, then we can merge this.

jordan-palmer:
I have nothing to add, this looks great!

PR #100, opened 2021-02-20
Simplify SpatialRateEnergySpectrum API

JelleAalbers:
This simplifies the API of the spatial rate sources. After this, users only have to specify a `spatial_hist` attribute, containing a regular, un-normalized, events/bin histogram of event positions. Flamedisx will do all normalization and conversions it needs internally.

---------

Currently, users instead have to 
  * convert their histograms to an undocumented normalization convention, and
  * pass the physical bin_volumes explicitly, even though we can deduce these from the histogram coordinate names and bin edges.
  
By trial/error/reverse-engineering, it seems we currently require something like this:
  * Convert histogram to a PDF (normalize, then divide by physical bin volume)
  * Divide by the mean over bins, __without__ weighting by any bin volume
  * Multiply by the bin volumes

Maybe there is a simpler way to satisfy the normalization requirement, but I've forgotten, and I can't find @pelssers' original notebook anymore. Flamedisx checks that the user input is correctly normalized according to the above convention, but it is easy to satisfy this check while messing up your distirbution (e.g. if you forget the physical bin volume correction in the PDF conversion, do some steps in the wrong order, etc.).

### DIFF
--- a/flamedisx/lxe_blocks/energy_spectrum.py
+++ b/flamedisx/lxe_blocks/energy_spectrum.py
@@ -173,61 +173,53 @@ class FixedShapeEnergySpectrum(EnergySpectrum):
 
 @export
 class SpatialRateEnergySpectrum(FixedShapeEnergySpectrum):
-    model_attributes = (('spatial_rate_hist', 'spatial_rate_bin_volumes')
+    model_attributes = (('spatial_hist',)
                         + FixedShapeEnergySpectrum.model_attributes)
     frozen_model_functions = ('energy_spectrum_rate_multiplier',)
 
-    spatial_rate_hist: Histdd
-    spatial_rate_bin_volumes: Histdd
+    spatial_hist: Histdd
 
     def setup(self):
-        # Check we actually have the histograms
-        for attribute, kind in [('spatial_rate_hist', Histdd),
-                                ('spatial_rate_bin_volumes', np.ndarray)]:
-            assert hasattr(self, attribute), f"{attribute} missing"
-            assert isinstance(getattr(self, attribute), kind), \
-                f"{attribute} should be a {kind}"
+        assert isinstance(self.spatial_hist, Histdd)
 
         # Are we Cartesian, polar, or in trouble?
-        axes = tuple(self.spatial_rate_hist.axis_names)
+        axes = tuple(self.spatial_hist.axis_names)
         self.polar = (axes == ('r', 'theta', 'z'))
-        if not self.polar:
+
+        self.bin_volumes = self.spatial_hist.bin_volumes()
+        if self.polar:
+            # Volume element in cylindrical coords = r * (dr dq dz)
+            self.bin_volumes *= self.spatial_hist.bin_centers('r')[:, None, None]
+        else:
             assert axes == ('x', 'y', 'z'), \
                 ("axis_names of spatial_rate_hist must be either "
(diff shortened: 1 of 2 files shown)


jordan-palmer:
These changes worked well for my application and is much more intuitive. Before and after the changes I return the correct source spatial distribution as what I put in to SpatialRateEnergySpectrum. Furthermore, the amount of effort from the notebook side has been minimized now, reducing the amount of histogram manipulation needed. An example is shown below for a generic NR source. First is the input histogram and second is the output of FD where the sum in each bin is forced to be the same. 
Screenshot 2021-02-23 at 09 34 44
Screenshot 2021-02-23 at 09 34 54

In terms of API, all I need to do now is the following. 
```
hist,edges=np.histogramdd([r,theta,z],weights=z,bins=(r_edges,theta_edges,z_edges))
### convert to multihist and make FD source
spatial_rate_hist = multihist.Histdd.from_histogram(hist,bin_edges=edges, axis_names=['r','theta','z']) #input into multihist

class spatialNR(fd.SpatialRateNRSource):
    spatial_hist=spatial_rate_hist
```
For an NR source, the next useful thing would be to have something like this as a function of energy. Im happy to begin looking into a new block for this

CristianAntochi:
Thanks for the update! I am looking at this as well for the Wall background case. I hope to come soon with an update

jordan-palmer:
Amazing! Maybe we should have a discussion on this topic to minimize duplication of work. I haven't started working on this yet however.

CristianAntochi:
Since we are here, discussing about the spatial dependence, I just found out a bug: when defining a spatial dependent source, it cannot have any "shape" parameter defined, otherwise the likelihood definition fails.
Say you have an er source
`class MyER( fd.SpatialRateERSource, fd.xenon.x1t_sr1.SR1ERSource):
      spatial_rate_hist = er_spatial_rate_hist/np.mean(er_spatial_rate_hist.histogram/er_spatial_rate_hist.bin_volumes())
      spatial_rate_bin_volumes=er_spatial_rate_hist.bin_volumes()

      @staticmethod
      def p_electron(nq, *, er_pel_a=4.24, er_pel_b=-80., er_pel_c=112.64, er_pel_e0=0.508): #4.24
      #(nq, *, er_pel_a=...., er_pel_b=-...., er_pel_c=...., er_pel_e0= ....):
          
          e_kev_sortof = nq * 13.7e-3
          eps = fd.tf_log10(e_kev_sortof / er_pel_e0 + 1e-9)
          qy = (
              er_pel_a * eps ** 2
              + er_pel_b * eps
              + er_pel_c)
          return fd.safe_p(qy * 13.7e-3)`

and you want to fit, say `er_pel_a`, when you construct the likelihood 

`ll = fd.LogLikelihood(
    sources=dict(wall=MyER),
    # Data to use:
    data=SR_data,
    # Parameters to fit, with (lower anchor, upper anchor, number of anchor points):
    er_pel_a=(-200., 60., 2),
    # Allow the absolute ER rate to be fitted as well
    free_rates=('wall',))`

You will get an error saying that 
`TypeError: draw_positions() got an unexpected keyword argument 'er_pel_a'`.

I will make a proper issue out of this comment

jordan-palmer:
I saw the same thing when trying to float e-lifetime but I wasn't sure if the issue was on my end (since I am seeing other issues!)

JelleAalbers:
It sounds like there are no objections to merging this?

Issue #101, opened 2021-02-20
Tensorflow error when calling to fd.LogLikelihood bestfit() function


sophiaandaloro:
I am currently fitting a source to data, and while it worked fine yesterday, as of today I get an error when running the LogLikelihood.bestfit function. Below is the code I am running (more or less) and the output error. Here's the main summary line: `OperatorNotAllowedInGraphError: using a tf.Tensor as a Python bool is not allowed: AutoGraph did convert this function. This might indicate you are trying to use an unsupported feature.`

 I have tried changing the optimizer, the dataset and the use_hessian arguments, but this seems to be a tensorfzlow issue? I am on flamedisx v1.3.0 and tensorfzlow v2.4.1, again no changes to any of this code since I ran yesterday. 
I'm happy to provide other information to help sort this out! 

Code:
```
fdll = fd.LogLikelihood(sources=dict(er=fd.SR1ERSource),
                                batch_size = 100,
                                free_rates = 'er',
                                data = rn_data_cut,
                                log_constraint = None, 
                        **kwargs # these are my fit variables, e.g. g1, gamma_er, omega_er., based on arxiv:1902.11297v2
                             ) 
fdll_guess = fdll.guess()
fdll_bestfit = fdll.bestfit(guess=fdll_guess, optimizer='scipy', use_hessian=True) #this is the breaking line
```

Output: 
```
---------------------------------------------------------------------------
OperatorNotAllowedInGraphError            Traceback (most recent call last)
<ipython-input-49-dafbd7a2e4bf> in <module>
      1 fdll_guess = fdll.guess()
----> 2 fdll_bestfit = fdll.bestfit(guess=fdll_guess, optimizer='scipy', use_hessian=True)

~/.conda/envs/sophia/lib/python3.7/site-packages/flamedisx-1.3.0-py3.7.egg/flamedisx/likelihood.py in bestfit(self, guess, fix, bounds, optimizer, get_lowlevel_result, get_history, use_hessian, return_errors, nan_val, optimizer_kwargs, allow_failure)
    513         # Check the likelihood has a finite value and gradient before starting
    514         val, grad, hess = self.log_likelihood(**guess,
--> 515                                               second_order=use_hessian)
    516         if not np.isfinite(val):
    517             raise ValueError("The likelihood is - infinity at your guess, "

[...]

PR #102, opened 2021-02-20
Name reconstruction efficiency pivots

JelleAalbers:
These are two small changes to x1t_sr1.py:
 * The bias and reconstruction efficiency pivot parameter names are currently `bias_pivot_pt1`, `bias_pivot_pt2` and `scalar`. These names end up in fit results, and have to be specified by the user (e.g. in the Likelihood construction, if you want to fit these parameters). This gives them a bit more descriptive names, `s1_reconstruction_bias_pivot`, `s2_reconstruction_bias_pivot` and `s1_reconstruction_efficiency_pivot`.
 * In `cal_rec_efficiency_tf`, an `if pivot_pt < 0` is used. Tensorflow _should_ be able to resolve this (see [here](https://www.tensorflow.org/guide/function#conditionals)), but it could be a problem for certain versions, or perhaps certain orders in which the functions are traced. Sophia ran into a problem with this in #101. This replaces the `if` with an explicit (but less readable) `tf.cond`, and renames `cal_rec_efficiency_tf` to `interpolate_reconstruction_efficiency`.

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -119,7 +119,7 @@ def interpolate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
     bias_low = interpolate_tf(sig_tf, fmap[0], domain_def)
     bias_high = interpolate_tf(sig_tf, fmap[1], domain_def)
 
-    bias = (bias_high-bias_low)*pivot_pt + bias_low
+    bias = (bias_high - bias_low) * pivot_pt + bias_low
     bias_out = bias + tf.ones_like(bias)
 
     return bias_out
@@ -292,7 +292,7 @@ class SR1Source:
 
 # ER Source for SR1
 @export
-class SR1ERSource(SR1Source,fd.ERSource):
+class SR1ERSource(SR1Source, fd.ERSource):
 
     @staticmethod
     def p_electron(nq, W=13.8e-3, mean_nexni=0.15,  q0=1.13, q1=0.47,
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -95,7 +95,7 @@ def interpolate_tf(sig_tf, fmap, domain):
             x_ref_min=domain[0], x_ref_max=domain[1], 
             y_ref=fmap, fill_value='constant_extension')
 
-def interpolate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
+def calculate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
     """ Computes the reconstruction bias mean given the pivot point.
 
     The pax reconstruction bias mean is a function of the S1 or S2 size and is
@@ -124,7 +124,7 @@ def interpolate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
 
     return bias_out
 
-def interpolate_reconstruction_efficiency(sig, fmap, domain_def, pivot_pt):
+def calculate_reconstruction_efficiency(sig, fmap, domain_def, pivot_pt):
     """ Computes the reconstruction efficiency given the pivot point
     :param sig: photon detected
     :param fmap: map returned by read_maps_tf
@@ -170,7 +170,7 @@ class SR1Source:
                                s1,
                                s1_reconstruction_bias_pivot=\
                                    DEFAULT_S1_RECONSTRUCTION_BIAS_PIVOT):
-        return interpolate_reconstruction_bias(
+        return calculate_reconstruction_bias(
             s1,
             self.recon_map_s1_tf,
             self.domain_def_s1,
@@ -180,7 +180,7 @@ class SR1Source:
                                s2,
                                s2_reconstruction_bias_pivot=\
                                    DEFAULT_S2_RECONSTRUCTION_BIAS_PIVOT):
-        return interpolate_reconstruction_bias(
+        return calculate_reconstruction_bias(
             s2,
             self.recon_map_s2_tf,
             self.domain_def_s2,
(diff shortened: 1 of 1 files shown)


flamedisx/xenon/x1t_sr1.py
```diff
@@ -95,7 +95,7 @@ def interpolate_tf(sig_tf, fmap, domain):
             x_ref_min=domain[0], x_ref_max=domain[1], 
             y_ref=fmap, fill_value='constant_extension')
 
-def cal_bias_tf(sig, fmap, domain_def, pivot_pt):
+def interpolate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
```
plt109, line 98: Minor comment but it would be better to call it `calculate_reconstruction_bias` to reflect its functionality. It carries out some computation instead of just interpolating from the bias maps.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -119,12 +119,12 @@ def cal_bias_tf(sig, fmap, domain_def, pivot_pt):
...
     bias_high = interpolate_tf(sig_tf, fmap[1], domain_def)
 
-    bias = (bias_high-bias_low)*pivot_pt + bias_low
+    bias = (bias_high - bias_low) * pivot_pt + bias_low
     bias_out = bias + tf.ones_like(bias)
 
     return bias_out
 
-def cal_rec_efficiency_tf(sig, fmap, domain_def, pivot_pt):
+def interpolate_reconstruction_efficiency(sig, fmap, domain_def, pivot_pt):
```
plt109, line 127: Similar minor comment to above, it would be better to call it `calculate_reconstruction_efficiency` to reflect its functionality. It carries out some computation instead of just interpolating from the bias maps.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -119,7 +119,7 @@ def interpolate_reconstruction_bias(sig, fmap, domain_def, pivot_pt):
     bias_low = interpolate_tf(sig_tf, fmap[0], domain_def)
     bias_high = interpolate_tf(sig_tf, fmap[1], domain_def)
 
-    bias = (bias_high-bias_low)*pivot_pt + bias_low
+    bias = (bias_high - bias_low) * pivot_pt + bias_low
```
plt109, line 122: Eggcellent! Seems like white space checker is working now.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -135,13 +135,11 @@ def cal_rec_efficiency_tf(sig, fmap, domain_def, pivot_pt):
...
     bias_median = interpolate_tf(sig_tf, fmap[1], domain_def)
 
-    if pivot_pt<0:
-        bias_other = interpolate_tf(sig_tf, fmap[0], domain_def)
-        bias_out = pivot_pt*(bias_median-bias_other)+bias_median
-    else:
-        bias_other = interpolate_tf(sig_tf, fmap[2], domain_def)
-        bias_out = pivot_pt*(bias_other-bias_median)+bias_median
-    return bias_out
+    bias_diff = tf.cond(
```
plt109, line 138: I considered using `tf.cond` while writing this (along with using `tf.math.multiply` instead of `*` and so on) just for explicitness. However, I decided against it in the end because the rest of flamedisx does not stick to strict explicitness when it comes to `tf` functions.

In this case, the 'if' statement is also used in other fundamental blocks that are traced over. (Just one example, but did not verify if there are more: https://github.com/FlamTeam/flamedisx/blob/e17ecdbfc841e53a82b21c7672698c16a49ba75b/flamedisx/lxe_blocks/quanta_splitting.py#L51) Replacing `if` with `tf.cond` might circumvent the issue at this point, but will be caught downstream in other traced functions in other blocks.

flamedisx/xenon/x1t_sr1.py
JelleAalbers: It's several interpolations in sequence: first you interpolate the different bias maps at the observed signals, then you interpolate the results for the chosen pivot. But happy to change it back :-)

flamedisx/xenon/x1t_sr1.py
JelleAalbers: The problem comes from the `if` statement being used on a tensor, whose value can change during the fit. There are of course many if statements in flamedisx, but as far as I can see, we always use `tf.where`, `tf.clip` etc. instead of `if` when dealing with tensors. There are several examples of this in `x1t_sr1.py`.

flamedisx/xenon/x1t_sr1.py
plt109: hahah yes i'm nitpicking but yes thanks!

there are 2 lines of interpolation, and then 2 lines of calculations from the interpolated results, not quite interpolating the results itself. https://github.com/FlamTeam/flamedisx/blob/a709e1747f0044e8cab95fb1fe5831b37586348d/flamedisx/xenon/x1t_sr1.py#L122

but yes. thanks!

flamedisx/xenon/x1t_sr1.py
plt109: Just for the sake of documentation so that I understand this issue in the future.

The issue is with using `if` on `pivot_pt` which is a tensor and its value can change during the process of fitting. The different between this and https://github.com/FlamTeam/flamedisx/blob/e17ecdbfc841e53a82b21c7672698c16a49ba75b/flamedisx/lxe_blocks/quanta_splitting.py#L51
is that `self.do_pel_fluct` is just a 'flag' and does not change during the fitting process.

plt109:
Thanks Jelle for the improvements! I think Cristian has made some other merges into master since, but please feel free to merge this PR into master.

PR #103, opened 2021-02-20
Minor likelihood fixes/enhancements

JelleAalbers:
A few small fixes/enhancements to flamedisx.Likelihood:
 * A new `defaults` argument for `Likelihood.__init__`. This allows you to override defaults for all sources. Now you can collect all your custom defaults in one place (e.g. a config file), load them into a dictionary, then pass them to your likelihood.
  * `Likelihood.guess` should output a dict of python values, not tf.constants. This is consistent with the output of e.g. `Likelihood.bestfit` and other functions.
 * If the likelihood had shape parameters, you could not call `.simulate()` without arguments. Now you can; omitted parameters are replaced by their defaults.
 * Conversely, if you passed shape parameters the likelihood did not take, those parameters were just ignored. So you might think   `Likelihood.simulate(elife=0)` runs a zero-elife simulation, but the elife was actually just left to its default value (unless the likelihood has `elife` as a shape parameter). Now you get an error if you try this. 

(We cannot just let `Likelihood.simulate` accept any parameters, like `Source.simulate`, since the former needs to know the efficiency at the new shape parameters. For this, the likelihood does a series of big simulations when it is initialized, a few for each parameter to be varied. We cannot just vary another parameter without doing more simulations. `Source.simulate` instead expects the user to say how many events are simulated, so it _can_ take arbitrary parameters. )

### DIFF
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -16,7 +16,7 @@ DEFAULT_DSETNAME = 'the_dataset'
 
 @export
 class LogLikelihood:
-    param_defaults: ty.Dict[str, float]
+    param_defaults: ty.Dict[str, tf.Tensor]
 
     dsetnames: ty.List[str]              # Dataset names
     sources: ty.Dict[str, fd.Source]     # Source name -> Source instance
@@ -464,9 +464,10 @@ class LogLikelihood:
                        0.)
         return ll
 
-    def guess(self):
+    def guess(self) -> ty.Dict[str, float]:
         """Return dictionary of parameter guesses"""
-        return self.param_defaults
+        return {k: v.numpy()
+                for k, v in self.param_defaults.items()}
 
     def params_to_dict(self, values):
         """Return parameter {name: value} dictionary"""
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -56,6 +56,7 @@ class LogLikelihood:
             log_constraint=None,
             bounds_specified=True,
             progress=True,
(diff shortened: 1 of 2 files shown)


flamedisx/likelihood.py
```diff
@@ -267,6 +274,7 @@ def set_data(self,
     def simulate(self, fix_truth=None, **params):
         """Simulate events from sources.
         """
+        params = self.prepare_params(params, free_all_rates=True)
```
CristianAntochi, line 277: Is this to clear the free rates param from the guess when simulating so you don't need to pop the free rate args from best fit when given to the simulation?

CristianAntochi:
Simple but effective changes! I think this can be merged without problems. @plt109 , thoughts?

flamedisx/likelihood.py
JelleAalbers: By default, only the rate multipliers of the sources in `free_rates` are legal parameters to the likelihood, and only those will appear in bestfit results. It can be useful for users of `.simulate` to be able to vary all rates, even those of fixed sources. If nothing else, we use this in the unit tests (https://github.com/FlamTeam/flamedisx/blob/master/tests/test_likelihood.py#L205). Not sure if this answers your question -- maybe you're thinking of a particular case where you are passing best-fit results to a simulation?

CristianAntochi: Ok, yes, I was mistakenly thinking of the bf simulation. I think I get it now. Thanks!

Issue #104, opened 2021-02-23
Cannot construct a likelihood dependent on a "shape parameter" if source has spatial rate dependency


CristianAntochi:
I just found out a bug: when defining a spatial dependent source, it cannot have any "shape" parameter defined, otherwise the likelihood definition fails.
Say you have an er source

```python
  class MyER( fd.SpatialRateERSource, fd.xenon.x1t_sr1.SR1ERSource):
        spatial_rate_hist = er_spatial_rate_hist/np.mean(er_spatial_rate_hist.histogram/er_spatial_rate_hist.bin_volumes())
        spatial_rate_bin_volumes=er_spatial_rate_hist.bin_volumes()


        @staticmethod
        def p_electron(nq, *, er_pel_a=..., er_pel_b=...., er_pel_c=..., er_pel_e0=...): #4.24
          
            e_kev_sortof = nq * 13.7e-3
            eps = fd.tf_log10(e_kev_sortof / er_pel_e0 + 1e-9)
            qy = (
                er_pel_a * eps ** 2
                + er_pel_b * eps
                + er_pel_c)
            return fd.safe_p(qy * 13.7e-3)
```

and you want to fit, say `er_pel_a`, when you construct the likelihood 
```python
ll = fd.LogLikelihood(
    sources=dict(wall=MyER),
    # Data to use:
    data=SR_data,
    # Parameters to fit, with (lower anchor, upper anchor, number of anchor points):
    er_pel_a=(-200., 60., 2),
    # Allow the absolute ER rate to be fitted as well
    free_rates=('er',))
```

You will get an error saying that 
`TypeError: draw_positions() got an unexpected keyword argument 'er_pel_a'`.

CristianAntochi:
Uhh, right! Easy fix!

JelleAalbers:
No worries; glad you found the bug!

PR #105, opened 2021-02-23
SpatialRateEnergySpectrum's draw_positions should accept parameters

JelleAalbers:
In flamedisx, `random_truth` and its minions `draw_positions` and `draw_time` get `**kwargs` of (shape) parameters. This is to help sources whose space , time, or energy distribution depend on fittable parameters. The default sources have no such dependencies, so they simply ignore these extra parameters.

`SpatialRateEnergySpectrum`'s `draw_positions` erroneously didn't accept this `**kwargs`, causing #104. This should fix that.


CristianAntochi:
Uhh, such an easy fix. I guess after playing all day with the spatial histograms and the modelling I wasn't thinking straight.

PR #107, opened 2021-03-03
Add recon bias

plt109:
The PAX reconstruction bias is defined as bias = (reconstructed_area - true_area)/true_area. This is implemented in this PR in `_simulate` and `_compute` of `MakeFinalSignals`. 

In `_simulate`, the reconstructed area is obtained by multiplying the bias to the true area that is drawn from the Normal distribution with the appropriate mean and standard deviations.

In `_compute`, the PDF of the events are evaluated at the true area instead of the reconstructed area.

`_annotate` is left untouched because it calculates bounds of the detected quanta(photoelectrons, electrons) which is not affected by the reconstruction bias.

### DIFF
--- a/flamedisx/lxe_blocks/final_signals.py
+++ b/flamedisx/lxe_blocks/final_signals.py
@@ -43,7 +43,7 @@ class MakeFinalSignals(fd.Block):
         tf.print('hi from _simulate')
         aa = self.gimme_numpy('reconstruction_bias_'+self.signal_name,
                 bonus_arg=d[self.signal_name])
-        d[self.signal_name] /= aa
+        d[self.signal_name] *= aa
         #### end test
 
         # Call add_extra_columns now, since s1 and s2 are known and derived
@@ -111,7 +111,7 @@ class MakeFinalSignals(fd.Block):
         # add offset to std to avoid NaNs from norm.pdf if std = 0
         result = tfp.distributions.Normal(
             loc=mean, scale=std + 1e-10
-        ).prob(s_observed*aa)
+        ).prob(s_observed/aa)
 
         # Add detection/selection efficiency
         result *= self.gimme(SIGNAL_NAMES[self.quanta_name] + '_acceptance',
--- a/flamedisx/lxe_blocks/final_signals.py
+++ b/flamedisx/lxe_blocks/final_signals.py
@@ -36,11 +36,9 @@ class MakeFinalSignals(fd.Block):
             scale=(d[self.quanta_name + 's_detected']**0.5
                    * self.gimme_numpy(self.quanta_name + '_gain_std')))
 
-        #### true area -> reconstructed area
-        aa = self.gimme_numpy('reconstruction_bias_'+self.signal_name,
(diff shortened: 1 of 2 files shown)


JelleAalbers (commented):
Thanks for the PR Pueh! In simulate, you do:

`observed = prebias * f(prebias)`

where `f` is the output of the `reconstruction_bias_` model function. Then in compute, you do:

`prebias = observed / f(observed)`

These cannot both be true unless `f(observed) = f(prebias)`, i.e. the bias is constant. Unless I'm mistaken, this would cause a discrepancy between the simulation and reconstruction wings of flamedisx.

The effect should usually be small, since the bias is usually small and changing slowly. Still, it could cause unpleasant surprises. Wouldn't you rather add a new Gaussian block after MakeFinalSignals, so you can also do reconstruction smearing?

Few minor comments:
  * Your comments say `reconstructed area = (bias+1)*true area` etc., but the bias model function as used here already includes the 1+, i.e. it's really a scaling rather than a bias.
  * Maybe it's clearer to speak of pre-bias area instead of true area? The observed area is 'true' in the sense that it is what was really seen.
  * I see a few PEP8 whitespace-related errors. Very soon these lint checks will be automated again... In the meantime, you could run the changed file through a linter yourself, e.g. `flake8`, to help you locate the problems.
  * This PR has many debug/fix commits. That's OK, but unless you'd like your "hi from _annotate but temporarily off" etc. to be preserved for all time, it can look good to squash such commits before making a PR. The easiest way is to just make a new branch, copy-paste the changes, and make one commit; you can find more advanced ways on Google. Alternatively, GitHub has an option to squash PRs into one commit when we merge, but it doesn't do so by default.

plt109:
Yes you are right that `f(observed) != f(prebias)`. Right now, `f=f(true_area)` and is retrieved from the BBF folder using `fd.get_bbf_file`. However, `f(observed)=f(reconstructed_area)` and this map currently does not exist, although I can create it from the original map and perhaps request it to be placed in the BBF folder(or put it somewhere else). Then `x1t_sr1` will have to load in 2 different maps for each direction. The difference between this map as a function of reconstructed_area and this map as a function of true_area is small but grows with the signal size so you are right again that it might pose a problem for large signals but right now the fits are still converging.

Ok, I agree that it is a better idea to restructure this in terms of a separate block after `MakeFinalSignals`. Currently `reconstruction_bias_s1` is a special model function of `MakeS1` which is why I thought I could just insert it there.

`reconstruction_bias_s1`, `reconstruction_bias_s2` calls `cal_bias_tf` which outputs `bias+1` so yes it is a scaling here because I thought it is better to wrap the `+1` inside `reconstruction_bias_s1` in case I accidentally leave it out later.

I would opt for the 'squash and merge' option on GitHub (in case I leave things out while copying and pasting), if you do not mind selecting the drop down option while merging eventually. But this will be in the very last step when the merging occurs.

JelleAalbers:
Superseded by https://github.com/FlamTeam/flamedisx/pull/273

PR #109, opened 2021-03-04
Fix automated tests, test with fixed dependencies

JelleAalbers:
Flamedisx's continuous integration testing is currently broken: TravisCI now requires payment, and CodeFactor seems to have gone silent  after https://github.com/FlamTeam/flamedisx/commit/8f231df9edb956bef03f44e00bb3f730819eef65 (and complained about too many things before that).

This will:
  * Add configuration for [GitHub Actions](https://docs.github.com/en/actions). This will run our unit tests and a code style checker (flake8) automatically on commits.
  * List one specific version of all dependencies in `requirements.txt`. Users have the option to use these version if they do `pip install flamedisx[strict-deps]`, but by default the version restrictions are unchanged (and now listed in `requirements_minimal.txt`). GitHub Actions will always use the strict/pinned dependency versions for testing; otherwise, the tests could randomly break if a dependency pushes an update.
  * Add configuration for [Dependabot](https://dependabot.com/), which (after this is merged and I change a few settings in the repo config) will make weekly PRs to update `requirements.txt` with new dependency releases, much like in https://github.com/XENONnT/base_environment/pulls. This will give us a chance to see if new tensorflow/numpy versions etc. break flamedisx or not.
  * Remove the now-obsolete codefactor and travis configurations.
  * Fix a few PEP8 errors, which may have crept into flamedisx while the style checker was off. I currently let flake8 ignore `x1t_sr1.py`: there are a few PEP8 things there, but I don't want to mess up the code Pueh is maintaining. @plt109, whenever you get a chance to look at that file, we can remove the ignore.

TravisCI also used to take care of uploading new releases to pypi. We'll have to do that manually, or add [another github action workflow](https://github.com/actions/starter-workflows/blob/618434daabf54b2f994b19c9e608c46db412c6a5/ci/python-publish.yml).


JelleAalbers:
Sorry guys, I am going to breach protocol and merge this; otherwise we can't really be sure any of the other PRs work.

PR #110, opened 2021-03-05
Fix 'sticky' defaults

JelleAalbers:
This fixes a bug in `Source._set_temporarily`. This low-level function is used during simulation and annotation to temporarily set data and parameters to different values. It's supposed to return a source to the previous state afterwards. 

... only it didn't properly reset the parameters. Thus, if you call simulate/annotate with non-default settings, the default settings actually get overwritten!

This can cause serious trouble, e.g. after

```python
import flamedisx as fd

ll = fd.LogLikelihood(
    sources=dict(er=fd.ERSource),
    elife=(1e6, 1e12, 2))
```
the `ll.sources['er'].defaults['elife']` is now actually 1e12 ns. If you now use e.g. `ll.sources['er'].simulate`, you will be in for a surprise... Fortunately, the likelihood does still remember the original defaults, so you're fine if you use `ll.simulate` instead. (In fact `ll.simulate` will reset the source's defaults to the original values.)

This is only one of the possible places this bug could have an impact though. I think we should merge this quickly -- and a new release is overdue anyway.

Thanks to @jordan-palmer for finding a clear case where this bug rears its ugly head.

### DIFF
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from contextlib import contextmanager
 import inspect
 
@@ -255,7 +256,7 @@ class Source:
         data tensor state"""
         if data is None:
             raise ValueError("No point in setting data = None temporarily")
-        old_defaults = self.defaults
+        old_defaults = deepcopy(self.defaults)
         if data is None:
             self.set_defaults(**kwargs)
         else:
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -1,4 +1,4 @@
-from copy import deepcopy
+from copy import copy
 from contextlib import contextmanager
 import inspect
 
@@ -256,7 +256,7 @@ class Source:
         data tensor state"""
         if data is None:
             raise ValueError("No point in setting data = None temporarily")
-        old_defaults = deepcopy(self.defaults)
+        old_defaults = copy(self.defaults)
         if data is None:
             self.set_defaults(**kwargs)
         else:
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -304,6 +304,11 @@ def test_set_data(xes: fd.ERSource):
         np.testing.assert_array_equal(xes.data['s1'], data2['s1'])
     np.testing.assert_array_equal(xes.data['s1'], data1['s1'])
 
+    # Setting defaults temporarily (see PR #110)
+    with xes._set_temporarily(data2, elife=100e3):
+        pass
+    assert xes.defaults['elife'] == fd.ERSource().defaults['elife']
+
     # Setting for real
     xes.set_data(data2)
     assert xes.data is not data1


CristianAntochi (approved):
Easy fix of a nasty bug!

jordan-palmer:
This was a really fast fix, thanks so much. Using set_data is just much more convenient and speedy :)

plt109:
So this bug appears when one does `ll.sources['er'].simulate`, but not `ll.simulate`? What about a custom source that inherits from `fd.SR1ERSource` and I call that custom source's simulate directly?

I have been doing 

`import flamedisx as fd

class mySource(fd.SR1ERSource):
    energies = tf.cast(tf.linspace(0., 13., 1000),
                       dtype=fd.float_type())

ll = fd.LogLikelihood(sources=dict(er=mySource, ac=myAC),
                                free_rates = 'er',
                                ...
                                elife = (1e6, 1e12, 2),
                             )

bf = ll.bestfit(guess=ll.guess(), optimizer='scipy')
some_mc = mySource(**bf).simulate(42)`

and I did not see anything out of the ordinary. Perhaps I have been wrong all along without realizing?

jordan-palmer:
So these are the examples I was using when I found this. For example, this method works 
```  df= fd.ERSource().simulate(100)
  ll = fd.LogLikelihood(
  sources=dict(er=fd.ERSource),
  data=df,
  elife=(100e3,800e3, 2), 
  free_rates=('er',))

  print(ll.bestfit(guess=dict(elife=450e3)))```
  
  However, this does not 
  
  ```  ll = fd.LogLikelihood(
  sources=dict(er=fd.ERSource),
  elife=(100e3,800e3, 2), 
  free_rates=('er',))
  
  s_er=ll.sources['er'].simulate(50)
  ll.set_data(s_er)
  print(ll.bestfit(guess=dict(elife=450e3)))
  
  ```
So I think your example is fine since you aren't using ll.source.simulate

**Comment by plt109** [comment]
Ah nice, thanks for the clarification!

So this bug only pops up when you call `Source.simulate` directly. It does not affect the `simulate` from any downstream classes that inherit from `fd.Source`.

[classes.lala.pdf](https://github.com/FlamTeam/flamedisx/files/6089810/classes.lala.pdf)

**Comment by jordan-palmer** [comment]
I think as long as you aren't using `ll.sources['er'].simulate(100)` you wont have an issue. From the original post from Jelle, it sounds like `ll.simulate(er_rate_multiplier=1)` will work as a replacement. I used 1 as an example here because I believe this will give 100 events.

**Comment by jordan-palmer** [comment]
One more thing, is it an issue that the elifetime fitted out is consistently too high, even when using 100 events? The lifetime is 452us and I am consistently fitting out 500/550us

**Comment by plt109** [comment]
My elife turns out on the high side as well. I was using ~3000 events. But in the long run, would you want to fit elife or input the measured values? I'm currently working on a feature that allows the user to read the elife from a measured values. It's largely sorted, I just haven't figured out an elegant method of referencing the elife lookup map/table because I reckon that different experiments would want to have a private repo of such numbers.

On a separate note, while testing this new branch, I bumped into some GPU ResourceExhaustedError (https://github.com/FlamTeam/flamedisx/issues/111). This occurs when the dataset has more than ~1000++ events. Do you observe the same thing while fitting a larger data set on the GPU?

**Comment by jordan-palmer** [comment]
This sounds good. For the LZ side of things, we are using a config file which you can access like 
```
import configparser
import os

config = configparser.ConfigParser(inline_comment_prefixes=';')
config.read(os.path.join(os.path.dirname(__file__), '../config', 'LZ_config.ini'))

elife=config.getfloat('DEFAULT','elife_guess')
```

We found this works nicely and can be experiment specific, there also isn't much of a speed decrease.

plt109:
Yeah it looks similar to what I have right now. https://github.com/FlamTeam/flamedisx/blob/b54ced0d276a4ae564f03c4642aeae232fb15fb3/flamedisx/xenon/x1t_sr1.py#L73

The GPU that I am using also has ~12 GB memory. I am currently floating 12 parameters but I would want to keep it as general as possible. The same fit converges on `master` though.

jordan-palmer:
I dropped you an email on something similar!

JelleAalbers:
OK, with the #111 mystery solved, let's merge this.

Issue #84, opened 2020-08-08
Better rescaling


JelleAalbers:
Currently flamedisx rescales inference parameters by dividing by the guess before passing our function to the optimizer. It would be better to subtract the guess, then divide by the distance between the right - left bound. See also #72.

JelleAalbers:
Fixed in #114. Some potential improvements are discussed there, but are more in the 'suggestions for further research' category than actionable issues.

PR #114, opened 2021-03-06
Better parameter normalization

JelleAalbers:
Flamedisx normalizes parameters before handing our likelihood to the optimizer. This helps some optimization methods converge. We currently normalize only by dividing out the (absolute value of the) guess, i.e. `x ->  x/ abs(guess)`, see #72, #76. 

This definitely helps, especially when parameters have similar relative uncertainties. However, as noted in #84, ideally you'd want to do `x -> (x - center)/spread`, so the likelihood will look similar in all parameters after normalization. 

This changes the normalization for bounded parameters to:
```
x -> (x - guess)/(right_bound - left_bound)
```
For unbounded parameters, the normalization is unchanged, i.e. we still simply divide by the guess.

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -130,15 +130,40 @@ class Objective:
             self._process_guess()
 
     def _process_guess(self):
-        """Part of initialization: perform ones self.guess and self.bounds
-        are set"""
+        """Final part of initialization, done after self.guess and self.bounds
+         are set"""
 
-        # We have the guess and bounds info, use them to define the scaled
-        # coordinates used internally by the optimizers.
-        # We always scale with positive numbers, so we don't have to reverse
-        # any bounds.
-        self.scale_vector = np.abs(self._dict_to_array(self.guess))
-        self.scale_vector[self.scale_vector == 0.] = 1.
+        # Use the guess and bounds to define normalized coordinates:
+        #     normalized = (physical - offset)/scale
+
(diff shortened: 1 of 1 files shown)


JelleAalbers:
I did some tests on a simple fit with a free er rate and two shape nuisance parameters: `g2`, and `er_pel_a`. I chose a deliberately mediocre guess. I tried 5 different datasets:
  * Nothing changes in how `TNC` behaves. Probably TNC applies its own normalization.
  * `trust-constr` (our default method) finishes faster with this PR; the number of iterations goes x0.5 - x0.8.
  * In 4 out of 5, `trust-constr` used to get stuck at the guess for `er_pel_a` on master. After this PR, it converges to a value consistent with the `TNC` result.

As an example, here is the fit history on dataset 3/5:



Here is the notebook, which includes the history plots for all five datasets: [test_scaling.zip](https://github.com/FlamTeam/flamedisx/files/6096486/test_scaling.zip)

I also tried setting the scaling based on the standard errors derived from the initial inverse Hessian. This has the advantage of also scaling unbound parameters (like the rate multipliers), but if the guess is weird, perhaps the scalings would be weird too. I didn't see much improvement in this example. Perhaps we could have this as an extra option later, or when we have a more serious fit convergence benchmark.

JelleAalbers:
We discussed this on slack, and Robert kindly had a look

PR #115, opened 2021-03-25
Read data from XENONnT Flamedisx private repo

plt109:
This allows the user to read files stored in the XENONnT/Flamedisx private repository after the necessary authentication token has been set up. 

```
fname_1t = '1t_maps/test_map.json'
map_1t = fd.get_nt_file(fname_1t)

fname_nt = 'nt_maps/test_map.json'
map_nt = fd.get_nt_file(fname_nt)
```

As the repository gets updated, the user might be required to update his/her local version of the repository. This can be achieved with

```
fd.refresh_nt()
```
which deletes the current local XENONnT Flamedisx private repo and clones the latest version.

### DIFF
--- a/flamedisx/xenon/data.py
+++ b/flamedisx/xenon/data.py
@@ -7,8 +7,6 @@ import os.path
 import flamedisx as fd
 export, __all__ = fd.exporter()
 
-import pdb
-
 
 # Path to the root folder of the BBF repository, if you have already cloned BBF
 # (Do not export it. If you do, reassigning fd.BBF_PATH won't update this one.)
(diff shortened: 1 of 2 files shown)


JelleAalbers (changes requested):
Hi Pueh, sounds good, though obviously I can't test it myself ;-)

Can you reduce the code duplication, i.e. factor out repeated code in functions? `ensure_nt` is literally a carbon copy of `ensure_bbf` with a different repository name, and `refresh_nt` is also very similar.

plt109:
Hello Jelle,
I have refactored out the bits that checks for the token into `ensure_token` and added a function that clones different repository according to the input arguments, `ensure_repo`. `refresh_nt` makes use of `ensure_token` as well.

JelleAalbers (approved):
Thank you Pueh, looks good to me!

Issue #90, opened 2020-08-20
Higher precision event times


pelssers:
When simulating event times we return uniform floats in between the source's `t_start` and `t_stop` datetimes. `t_start.value` and `t_stop.value` are big integers of order `1e18`. Event times in data are also integers of the same magnitude and have (10x)nanosecond precision. Using floats for the simulated event times is inconsistent but also less precise. In particular it loses at most `2**7=128 ns` when using float64 and at most `2**36~69 s` when using float32.
This is not a problem for flamedisx at the moment since the time dependence in our models is on much larger time scales. But adding things like cut acceptances might depend on runs which are of order 1h so it might become a problem in the future.

Drawing integers instead of floats is a trivial change in `EnergySpectrum.draw_time`, but we are using float32 in most of the internal calculations as well and should check where exactly event_times are used and where the precision is needed.

JelleAalbers:
#99 is a workaround for the CoordinateOutOfRange errors (see also #98). However, as Bart writes, this issue will still come up when using run-dependent quantities.

PR #118, opened 2021-04-01
Elife interpolated from elife map from private repo

plt109:
Changes only affect the ER model in xenon/x1t_sr1.py

The electron lifetime is interpolated from the electron lifetime map pulled from the XENONnT/Flamedisx private repository based on the event time of the event and added as a column in `add_extra_columns`. Subsequently, also in `add_extra_columns`, the cs2 value is calculated using this interpolated electron lifetime instead of the default value of 641us which was previously hardcoded in.

The electron lifetime information is then used in model function `electron_detection_eff` as a positional argument. It was previously a keyword argument (nuisance parameter) which had a default value of 641us.

Performance of this new model, along with ks p-value distributions from multiple MC toys, is documented in [Convo note](http://links.app.convo.com/v1/acc-d87cde0a-409c-11e3-adfe-12313d0421bc/apps/6/messages/Y7CezL6ZgDwU7xOzATirnC76NGL1k8#title=Elife%252520bug%25253A%252520now%252520s2%252520fits%252520well%252520but%252520not%252520cs2%25253F) (under flamedisx tag).

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -20,7 +20,6 @@ DEFAULT_AREA_FRACTION_TOP = 0.63  # fraction of light from top array
 DEFAULT_P_DPE = 0.219
 DEFAULT_EXTRACTION_EFFICIENCY = 0.96
 
-DEFAULT_ELECTRON_LIFETIME = 641e3
 DEFAULT_DRIFT_VELOCITY = 1.34 * 1e-4   # cm/ns, from analysis paper II
 
 DEFAULT_DRIFT_FIELD = 81.
@@ -215,9 +214,8 @@ class SR1Source:
                           d['y'].values,
                           d['z'].values]))
 
-        # Not sure what this old comment is about, did Jordan's patch fix this?
-        # Not good. patchy. event_time should be int since event_time in actual
-        # data is int in ns
+        # Not too good. patchy. event_time should be int since event_time in actual
+        # data is int64 in ns. But need this to be float32 to interpolate.
         if 'elife' not in d.columns:
             d['event_time'] = d['event_time'].astype('float32')
             d['elife'] = interpolate_tf(d['event_time'], self.elife_tf[0],
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -63,6 +63,11 @@ path_reconstruction_efficiencies_s1 = ['RecEfficiencyLowers_SR1_70phd_v1.json',
 path_cut_accept_s1 = ['S1AcceptanceSR1_v7_Median.json']
 path_cut_accept_s2 = ['S2AcceptanceSR1_v7_Median.json']
 
+##
+# Loading elife maps
+##
+path_electron_lifetimes = ['1t_maps/electron_lifetimes_sr1.json']
+
 def read_maps_tf(path_bag, is_bbf=False):
     """ Function to read reconstruction bias/combined cut acceptances/dummy maps.
     Note that this implementation fundamentally assumes upper and lower bounds
@@ -77,8 +82,7 @@ def read_maps_tf(path_bag, is_bbf=False):
         if is_bbf:
             tmp = fd.get_bbf_file(loc_path)
         else:
-            with open(loc_path) as json_file:
-                tmp = json.load(json_file)
+            tmp = fd.get_nt_file(loc_path)
         yy_ref_bag.append(tf.convert_to_tensor(tmp['map'], dtype=fd.float_type()))
         data_bag.append(tmp)
     domain_def = tmp['coordinate_system'][0][1]
@@ -166,6 +170,10 @@ class SR1Source:
         self.recon_map_s2_tf, self.domain_def_s2 = \
             read_maps_tf(path_reconstruction_bias_mean_s2, is_bbf=True)
 
+        # Loading electron lifetime map
+        self.elife_tf, self.domain_def_elife = \
+            read_maps_tf(path_electron_lifetimes, is_bbf=False)
+
     def reconstruction_bias_s1(self,
                                s1,
                                s1_reconstruction_bias_pivot=\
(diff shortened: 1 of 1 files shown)


JelleAalbers:
@CristianAntochi, since this only affects the XENON models, would you like to take a look?

plt109:
actually this is just 1T model and you can see the convo note as well cause I attached the flamedisx tag as well.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -207,6 +214,13 @@ def add_extra_columns(self, d):
                           d['y'].values,
                           d['z'].values]))
 
+        # Not too good. patchy. event_time should be int since event_time in actual
+        # data is int64 in ns. But need this to be float32 to interpolate.
+        if 'elife' not in d.columns:
+            d['event_time'] = d['event_time'].astype('float32')
```
CristianAntochi, line 220: Maybe double precision would be necessary here since we deal with interpolations of ns

flamedisx/xenon/x1t_sr1.py
```diff
@@ -207,6 +214,13 @@ def add_extra_columns(self, d):
...
+            d['elife'] = interpolate_tf(d['event_time'], self.elife_tf[0],
+                                    self.domain_def_elife)
+
```
CristianAntochi, line 223: empty line

flamedisx/xenon/x1t_sr1.py
```diff
@@ -216,14 +230,14 @@ def add_extra_columns(self, d):
             d['cs2'] = (
                 d['s2']
                 / d['s2_relative_ly']
-                * np.exp(d['drift_time'] / self.defaults['elife']))
+                * np.exp(d['drift_time'] / d['elife']))
+
```
CristianAntochi, line 234: extra empty line

flamedisx/xenon/x1t_sr1.py
```diff
@@ -20,7 +20,6 @@
 DEFAULT_P_DPE = 0.219
 DEFAULT_EXTRACTION_EFFICIENCY = 0.96
 
-DEFAULT_ELECTRON_LIFETIME = 641e3
```
CristianAntochi, line 23: We should leave the constant lifetime as an option

flamedisx/xenon/x1t_sr1.py
```diff
@@ -216,14 +230,14 @@ def add_extra_columns(self, d):
             d['cs2'] = (
                 d['s2']
                 / d['s2_relative_ly']
-                * np.exp(d['drift_time'] / self.defaults['elife']))
+                * np.exp(d['drift_time'] / d['elife']))
+
 
     @staticmethod
     def electron_detection_eff(drift_time,
+                               elife,
```
CristianAntochi, line 238: As before, we should have the option for constant elife as well

CristianAntochi (changes requested):
Looks good, but I would prefer to let in the possibility to have constant lifetime as well, just in case.

flamedisx/xenon/x1t_sr1.py
plt109: ok. i've reintroduced the option of having constant elife.

flamedisx/xenon/x1t_sr1.py
plt109: I have reintroduced the option of having constant elife using the variable `variable_elife` which will set elife to a constant in `add_extra_columns` (see https://github.com/FlamTeam/flamedisx/blob/d620b0fddba6612f2b67476c086b046ecf9d5b3b/flamedisx/xenon/x1t_sr1.py#L229).

However, it is not possible to let elife float as a nuisance parameter now. But according to [the convo note](http://links.app.convo.com/v1/acc-d87cde0a-409c-11e3-adfe-12313d0421bc/apps/6/messages/Y7CezL6ZgDwU7xOzATirnC76NGL1k8#title=Elife%252520bug%25253A%252520now%252520s2%252520fits%252520well%252520but%252520not%252520cs2%25253F) (which Jelle should be able to see well), keeping it fixed at 641us (analysis paper I value) gives a better fit than letting it float as a nuisance parameter.

flamedisx/xenon/x1t_sr1.py
plt109: Good suggestion. I gave it some thoughts and realised that keeping it as float32 might be better. This is vaguely related to https://github.com/FlamTeam/flamedisx/issues/90 about choosing between int and float32, but the option of using float64 instead float32 does not seem to present much advantages.

The precision is constrained by the elife map from MongoDB. The timestamps are the UNIX time in s, not ns, even though they are stored as float64.

In this PR, the elife maps are read in using `read_maps_tf` as `fd.float_type()` (see https://github.com/FlamTeam/flamedisx/blob/b9c98fbae154a055f3527d017bd05c8a9aa8b1ab/flamedisx/xenon/x1t_sr1.py#L89), where `fd.float_type` is defined as `tf.float32` (see https://github.com/FlamTeam/flamedisx/blob/a338ab35b9ce0fc1649620dd3de113021468caa8/flamedisx/utils.py#L41), which is why `d['event_time']` is required to be float32.

As mentioned by Bart in https://github.com/FlamTeam/flamedisx/issues/90, most of the internal computations are in float32, though there are some exceptions, especially in `likelihood.py`.

CristianAntochi: Yeah I agree. The thing that made me a bit worry is the fact that you already lose some info going from int64 to float64 when dealing with large numbers (millenium bug, somehow related). In float32 you definitely loose nanosecond precision if data is in int64(ns). I guess the comment can be misleading. Anyway in case we ever need a proper interpolation and we see weird things this is a thing to be aware of.

CristianAntochi (commented):
Good that we don't let the elife float. The elife map should be a standard implementation for XENON use, but for external use we should give some directives in how to treat the elife, what functions to modify if one wants to change the elife map or even create e function that gives elifes at different times. So docstrings. We need to be more careful with the documentation. This is a general comment, and reviewing the code, it doesn't need more infos, as is.

CristianAntochi (approved):

flamedisx/xenon/x1t_sr1.py
plt109: ok. i'll just reference https://github.com/FlamTeam/flamedisx/issues/90 in the comments instead.

PR #127, opened 2021-05-04
Rj variable stepping

robertsjames:
Summary [here ](https://docs.google.com/presentation/d/1b4xh2-nUhBhnscXNoUEcthufW74W4wGHXgzD52QBe4Q/edit?usp=sharing)

This PR implements variable stepping into Flamedisx, for both the current default models and any future sets of models (i.e. the NEST models which I will port over next).

This was a LOT more work than I had envisaged: this is due to the way that the FD default models relate the energy spectrum to the quanta tensor, and the associated issues that variable stepping caused there. It wasn't an issue present in the NEST version, so I hadn't had to solve it until now. 

The slides should summarise this, the solution, the implementation of the stepping and the various other features I had to implement to get around this problem. I have tried to keep those features very generalised so they may be of use to others in the future!

### DIFF
--- a/flamedisx/lxe_blocks/quanta_generation.py
+++ b/flamedisx/lxe_blocks/quanta_generation.py
@@ -40,10 +40,12 @@ class MakeERQuanta(fd.Block):
         result = tf.cast(tf.equal(quanta_produced_noStep,
         quanta_produced_real), dtype=fd.float_type())
 
-        result_pad_left = tf.pad(result, [[0, 0],
-        [15 - result.shape[1] % 15, 0], [0, 0]])
-        result_pad_right = tf.pad(result, [[0, 0],
-        [0, 15 - result.shape[1] % 15], [0, 0]])
+        padding = int(np.floor(quanta_produced_noStep.shape[1] / \
+        (quanta_produced.shape[1]-1))) - \
+        quanta_produced_noStep.shape[1] % (quanta_produced.shape[1]-1)
+
+        result_pad_left = tf.pad(result, [[0, 0], [padding, 0], [0, 0]])
+        result_pad_right = tf.pad(result, [[0, 0], [0, padding], [0, 0]])
 
         steps = self.source._fetch('quanta_produced_steps',
         data_tensor=data_tensor)
--- a/flamedisx/lxe_blocks/quanta_generation.py
+++ b/flamedisx/lxe_blocks/quanta_generation.py
@@ -89,6 +89,7 @@ class MakeERQuanta(fd.Block):
 class MakeNRQuanta(fd.Block):
 
     dimensions = ('quanta_produced', 'energy')
+    extra_dimensions = (('quanta_produced_noStep', False),)
     depends_on = ((('energy',), 'rate_vs_energy'),)
 
     special_model_functions = ('lindhard_l',)
@@ -116,17 +117,47 @@ class MakeNRQuanta(fd.Block):
                  # Domain
                  quanta_produced,
                  # Dependency domain and value
-                 energy, rate_vs_energy):
+                 energy, rate_vs_energy,
+                 # Extra tensors for internal use
+                 quanta_produced_noStep, energy_noStep):
 
         work = self.gimme('work', data_tensor=data_tensor, ptensor=ptensor)
         mean_q_produced = (
-                energy
-                * self.gimme('lindhard_l', bonus_arg=energy,
+                energy_noStep
+                * self.gimme('lindhard_l', bonus_arg=energy_noStep,
                              data_tensor=data_tensor, ptensor=ptensor)
                 / work[:, o, o])
 
         # (n_events, |nq|, |ne|) tensor giving p(nq | e)
-        return tfp.distributions.Poisson(mean_q_produced).prob(quanta_produced)
+        result = tfp.distributions.Poisson(mean_q_produced).prob(
+        quanta_produced_noStep)
+
+        padding = int(tf.floor(tf.shape(quanta_produced_noStep)[1] / \
+        (tf.shape(quanta_produced)[1]-1))) - \
+        tf.shape(quanta_produced_noStep)[1] % (tf.shape(quanta_produced)[1]-1)
+
+        result_pad_left = tf.pad(result, [[0, 0], [padding, 0], [0, 0]])
+        result_pad_right = tf.pad(result, [[0, 0], [0, padding], [0, 0]])
+
+        steps = self.source._fetch('quanta_produced_steps',
+        data_tensor=data_tensor)
+        chunks = int(tf.shape(result_pad_left)[1] / tf.shape(quanta_produced)[1])
+
+        result_temp_left = tf.reshape(result_pad_left,
+        [tf.shape(result_pad_left)[0],
+        int(tf.shape(result_pad_left)[1] / chunks), chunks,
+        tf.shape(result_pad_left)[2]])
+        result_left = tf.reduce_sum(result_temp_left, axis=2) \
+        / (steps[:, o ,o] * steps[:, o ,o])
+
+        result_temp_right = tf.reshape(result_pad_right,
+        [tf.shape(result_pad_right)[0],
+        int(tf.shape(result_pad_right)[1] / chunks), chunks,
+        tf.shape(result_pad_right)[2]])
+        result_right = tf.reduce_sum(result_temp_right, axis=2) \
+        / (steps[:, o ,o] * steps[:, o ,o])
+
+        return (result_left + result_right) / 2
 
     def _simulate(self, d):
         # If you forget the .values here, you may get a Python core dump...
@@ -136,7 +167,16 @@ class MakeNRQuanta(fd.Block):
         d['quanta_produced'] = stats.poisson.rvs(energies * lindhard_l / work)
 
     def _annotate(self, d):
+        d['quanta_produced_noStep_min'] = (
+                d['electrons_produced_min']
+                + d['photons_produced_min'])
         annotate_ces(self, d)
+
+    def _domain_dict_bonus(self, d):
+        return domain_dict_bonus(self, d)
+
+    def _calculate_dimsizes_special(self):
+        return calculate_dimsizes_special(self)
 
 
 def annotate_ces(self, d):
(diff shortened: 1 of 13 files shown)


flamedisx/source.py
```diff
@@ -31,6 +31,12 @@ class Source:
     # Final observable dimensions; for use in domain / cross-domain
     final_dimensions = tuple()
 
+    # Avoid variable stepping over these inner_dimensions
+    no_step_dimensions = tuple()
+
+    # Any non-hidden variable extra_dimensions
+    bonus_dimensions = tuple()
```
sophiaandaloro, line 38: @robertsjames  Should this be `extra_dimensions` now? I see that you renamed `bonus_dimensions` to `extra_dimensions` in the block source codes (or is `bonus_dimensions` the general name, and `extra_dimensions` a specific use of it??)

robertsjames: `extra_dimensions` is now the generic name for any additional dimensions you want a block to see. If they represent an internally contracted hidden variable, they will get added to `inner_dimensions` (this will be the case for the ions in the NEST implementation). Otherwise they get added to `bonus_dimensions`, where you can do what you like with them (this is the case here for quanta_produced_noStep).

JelleAalbers:
Thanks for this Robert! Variable stepping is essential for having flamedisx deal with higher-energy events, and achieving better speed in the upcoming NEST models with more complicated inner dimensions -- so I'm very happy to see this arrive.

I'm trying to understand the general idea first. Let's definitely go through the details afterwards (and once you have made the test pass, of course :smiling_imp:).

I'll summarize what I'm understanding so far below; just correct me where I'm wrong.

Summary
--------
The main idea is to take fixed-size steps (but allow different step sizes for different events) in large dimensions, do the computations as usual, and finally multiply the differential rate results with the step size used.

In more detail:
  * First, whenever data is set, [this](https://github.com/robertsjames/flamedisx/blob/RJ-variable_stepping/flamedisx/source.py#L277) assigns each _event_ a personal step size for each inner dimension in a block model. 
     * The step size is `ceil[(size - 1) / (cap - 1)]`, where `size` is the original dimension size (`max - min + 1`) estimated by the bounds estimation, and `cap` is the user-configured dimension size cap.
  * `Source.domain` uses these step sizes [here](https://github.com/robertsjames/flamedisx/blob/RJ-variable_stepping/flamedisx/block_source.py#L300) to calculate the allowed domain values for each inner dimension. 
    * Specifically, the domain is [`min`, `min + step`, ... `min + N * step`], with N the largest dimension size among events in one batch. 
    * Unlike master, domain shapes differ from batch to batch (and thus block result shapes also).
  * Next, most blocks are in computed as usual (some blocks need special care -- see below);
  * Finally, [this](https://github.com/robertsjames/flamedisx/blob/RJ-variable_stepping/flamedisx/block_source.py#L300) applies the weighting for each step, i.e. multiplies the differential rates of a blocks using stepped dimensions by the stepsize.
    * Each dimension only causes one stepsize multiplication, even if multiple blocks output results using that dimension.

Quanta tensor problem?
----------------------

You wrote that there are problems relating to the (photons_produced, electrons_produced) block, i.e. the computation in `MakePhotonsElectronsBinomial`.

I can definitely see why this block is problematic (it always is :wink:). `MakePhotonsElectronsBinomial` uses the result of another block (the `MakeERQuanta` or `MakeNRQuanta` result) in a non-standard way: it looks up the differential rate for `n_quanta = (n_photons + n_electrons)` in the other block's output.

I may not be fully grasping the implications of this for stepping though. I would guess there are two:
1. If the n_quanta dimension uses stepping, non-existent n_quanta entries may be queried.
2. If photons or electrons are stepped, the `quanta = photons + electrons` relation results in biased lookups. For example, moving from step size 1 to 2 replaces four queries of (0, 1, 1, 2) quanta in the blue square below by one (4x weighted) query of 0 quanta produced:

stepping

(I don't see a normalization issue though, as long as we weigh the stride-2 calculation by 4x, which it looks like you do. There is no problem from 'missing' the entries for 1 and 3 generated quanta: their neighbours get upweighted as a result.)

If those are indeed the problems, I would try to 
  * To solve 1, either force the `MakeERQuanta`/`MakeNRQuanta` result to use step size 1 (if that is still OK speed-wise) or round lookups to the nearest existing quanta_produced entry (which could get tricky: it would probably involve int<->float conversions, which caused speed problems involving CPU<->GPU copying for us in the past). Looks like you did the first, and it was indeed OK speed-wise.
  * To solve 2, I would look `n_photons + n_electrons + (photon_step_size + electron_step_size)//2` quanta instead of `n_photons + n_electrons`. (Or if we implemented nearest-existing lookups for 1, we could perhaps replace // by /, and have the nearest-existing lookup choose the smarter way to round).
 
You did something more intricate, which I have trouble pushing my brain to grok until I understand why this simpler solution wouldn't work / is based on an incomplete understanding of the issue. :-)

Other question
---------------
  * If tensors change size each batch, does that mean tensorflow retraces the graph for every possible batch size? If so, it might cause a large memory use and slow setup time, especially in analyses that run over multiple datasets (e.g. sensitivity checks or Wilks validations).
  * Using a fixed step size per event leads to larger relative strides at low values, e.g. stride length 2 means we go from 100 to 102 skipping 101 (fine), but also from 0 to 2 while skipping 1 (maybe not fine).
  
It seems both these issues would be avoided if we used uniform relative / logarithmic step sizes. For reweighting, you can recover the step sizes from the domains / domain_dict, which would also eliminate the `dim + '_steps'` variables. There is the annoying fence point issue of course (one less step size than items in a domain), but just repeating the final step size once more seems harmless if the dimension size cap is large enough.

JelleAalbers:
As Robert explained to me, I was mistaken when I wrote "There is no problem from 'missing' the entries for 1 and 3 generated quanta: their neighbours get upweighted as a result". What if a source can only make 2 generated quanta? Then upweighting neighbours is futile, since they are zero! More generally, naive stepping with lookups will be inaccurate near sharp features in the quanta_produced spectrum.

Similar problems could happen everywhere due to stepping: if a detector is blind to exactly 42-electron S2s, but not 41 or 43-electron S2s, stepping in electrons_detected may also cause flamedisx to miss that. But such sharp features a less likely for variables further away from the original spectrum. Produced quanta is still very close, so we have to pay extra attention it.

To improve on lookups, we can do weighted averages over several `quanta_produced`. But if each cell in the quanta tensor would need its own private weighted-average operation, that would be expensive. (A double for loop over weighted average, or constructing and contracing with a (1+3)d weight tensor `(n_events, |electrons_produced|, |photons_produced|, |quanta_produced|)`) And if it is expensive, there would no point: stepping is only an optimization.

Robert's solution has two parts:
  * Enforce that photons and electrons produced use the same, fixed, integer step size (for all events in a batch). Then all "lookups" go to a fixed grid of `quanta_produced` values, and the computation to do is always the same every time we "look up" one particular `quanta_produced`. 
  * Hence, we can precompute the weighted averages and store them in a (1x1)d tensor, which will be a (n_events, |quanta_produced|) differential rate re-weighted to a different `quanta_produced` domain. We can now just do lookups into that tensor.

This really is quite clever, I have no idea how to solve this efficiently for arbitrary stepping schemes (e.g. #133). Maybe we'll think of something later, or we eventually find a more accurate weighted average scheme than used here -- but clearly those are extra complications, and we should be merging this first. 

About using internal tensors that vary in shape between batches: Robert mentioned tensorflow does not retrace for each batch, and instead makes one trace where some internal tensors have extra `None` dimensions. The input signature of our `tf.function` is not changed by making internal tensors variable-size (unlike e.g. if we'd make the number of events per batch variable). It would still be worth testing the performance on GPUs, ideally with different batch sizes, but this could very well be no problem or even make things faster.

robertsjames:
Thanks for the nice summary of our chat yesterday @JelleAalbers! I'm glad you understand the problem now, and appreciate my solution :D

I will sort the CI out and confirm that GPU speed is still okay with variable tensor sizes in the next day or two. Then we can get this thing merged!

robertsjames:
Confirmed that the variable tensor size feature causes (next to) no slowdown on GPU:

47a046e52318582b9fa373f5c50cbdb8c778a971

fe39020bc0fdc7233be4fc1b3d491f522bcc1390

Top image is 4 rate calculations (10,000 events, batch size 10, 0-10 keV ER source, so negligible impact of stepping) timed on a Colab session on commit 47a046e52318582b9fa373f5c50cbdb8c778a971 (most recent in the PR), bottom image is the same on commit fe39020bc0fdc7233be4fc1b3d491f522bcc1390 (before any of the commits in this PR).

Unsure whether the slight change is related to the variable tensor sizes or not; either way, I am sure we can all agree that the order of magnitude speed improvement one gets with variable stepping at higher energies far offsets this!!

JelleAalbers:
Nice! And with larger batch sizes (with GPUs we're usually in the 100s) it should be even better.

flamedisx/lxe_blocks/quanta_generation.py
```diff
@@ -117,3 +237,83 @@ def annotate_ces(self, d):
...
+        + (d['photons_produced_steps'] <
+           d['electrons_produced_steps']) * d['photons_produced_steps']
+
+    batch_size = self.source.batch_size
+    n_batches = self.source.n_batches
+
+    # Need the electrons/photons steps to be the same within a batch for the
+    # averaging procedure in _compute to work correctly
+    for i in range(n_batches):
+        quanta_steps[i * batch_size: (i + 1) * batch_size + 1] = \
```
JelleAalbers, line 280: Super minor comment, but you could assign `i * batch_size` and `(i + 1) * batch_size + 1` to variables rather than computing them twice. To avoid repeating these lines later, you could factor it out into a function (`_unify_batch_steps` or something).

flamedisx/lxe_blocks/quanta_generation.py
```diff
@@ -117,3 +237,83 @@ def annotate_ces(self, d):
...
+        energy_domain[:, o, :],
+        tf.shape(quanta_produced_noStep_domain)[1],
+        axis=1)
+
+    # Return as domain_dict
+    return dict({'quanta_produced_noStep': quanta_produced_noStep,
+                 'energy_noStep': energy_noStep})
+
+
+def calculate_dimsizes_special(self):
```
JelleAalbers, line 263: Could you add a docstring or multiline comment outlining the steps you go through here?

JelleAalbers (commented):
Thanks Robert, this looks great! I added a couple of questions / suggestions, hope you can take a look at those. The most substantial point is about `extra_dimensions` / `bonus_dimensions` and how Blocks should set those.

How did you choose the default domain size of 70? Is this still fine for the low-energy applications of flamedisx? You could also consider increasing the default or even put 999_999 until we do more studies.

PR #128, opened 2021-05-05
removed pdb and extra unused inputs for p_electron

plt109:
Removed `import pdb as pdb` and unused inputs,`r, z`  for `p_electron`.

These lines were missed out during the review of PR #118.


CristianAntochi (approved):
Sorry for not spotting this last time. I think this should be merged ASAP

PR #136, opened 2021-06-04
Rj nest post quanta fixed

robertsjames:
Summary [here](https://docs.google.com/presentation/d/17RqGsUMItP2yu3iLfGlZJ02z0JWuR7PDRdOruw-sr3s/edit)

This PR implements the full set of NEST models after quanta production/splitting occurs (i.e. detector response) for NEST's S1 useTiming=-1 option and S2 useTiming=0 option (ask me or @sophiaandaloro if you don't know what this means). Pre-quanta models will follow shortly in another PR, and hopefully the S1 useTiming=0 option will follow after that.

Made this a new PR to fix some problems with conflicts, now that I have merged #127.


flamedisx/likelihood.py
```diff
@@ -51,8 +51,8 @@ def __init__(
                 ty.Dict[str, pd.DataFrame]] = None,
             free_rates=None,
             batch_size=10,
-            max_sigma=3,
-            max_dim_size=70,
+            max_sigma=12,
```
sophiafarrell, line 54: Sorry I am forgetting this @robertsjames why does max sigma need to be so large (12)?

sophiafarrell (commented):
This is phenomenal and I think helps streamline the "building" of one's detector in flamedisx in general. 
I am happy to see checks are passing and will look into those next, but the code looks great and I believe only changes made were to get checks to pass and all physics has been verified on my end to match NEST at the post-quanta level.

sophiafarrell:
Yep looking at the tests, it shows our default tests are still passing which is great. Should we also ensure that files in the directory `tests` now also test on the nest blocks, as well? Would probably double (or more?) computing time for tests, but perhaps we don't need to do all source tests at this point, and can just focus on say the ER test.

flamedisx/likelihood.py
jordan-palmer: I think it does until the bounds are updated

robertsjames: This is a result of adding in the S1/S2 noise, which makes the MLE of the final detected quanta subject to more fluctuations. This is back-propagated through the earlier bounds estimations and so they end up not covering enough without us making max_sigma larger. With any luck the MC bounds (coming next) will solve this problem...

robertsjames:
Hey @sophiafarrell, glad you're liking the look of this! I agree we want some CI tests for the NEST sources (and the variable stepping). Perhaps this is something you'd like to work on? Otherwise I will get to it once everything else is done

sophiafarrell:
Definitely! Looking over it this week, I didn't realize that already Jelle's PR over at https://github.com/robertsjames/flamedisx/pull/1 would already include these nice suggestions! Though the tests are basic, this is exactly what I had in mind at this point, just something to catch any "broken" issues. We can definitely sophisticate this further, but I think Jelle had already the test suite idea I was suggesting (thanks @JelleAalbers!)

JelleAalbers:
I added some suggestions in robertsjames#1 last week, once we resolve those I'd be happy to merge this. My main comment is to undo a few renamings and keep the nest sources into a subfolder/package, so our existing flamedisx notebooks can keep working.

There is inevitably quite some duplication between the NEST and regular sources, but I'd propose we address that after all the NEST stuff is in. I don't want to make Robert's porting work any harder than it already is!

robertsjames:
I think this is ready to be merged into master following my merging of Jelle's suggestions?

JelleAalbers (approved):
All looks good to me, thanks Robert! Another big step on the NEST journey.

PR #138, opened 2021-06-17
fix s2 acceptance and force float in map reading

CristianAntochi:
Small pull request to fix the SR1 S2 acceptance to account for the S2<200pe necessary for sources that produce low S2s and force the map reading to read floats.

### DIFF
--- a/flamedisx/xenon/itp_map.py
+++ b/flamedisx/xenon/itp_map.py
@@ -28,7 +28,7 @@ class InterpolateAndExtrapolate:
         averaging. Default is 2 * dimensions of points.
         """
         self.kdtree = cKDTree(points)
-        self.values = values
+        self.values = values.astype(np.float)
         if neighbours_to_use is None:
             neighbours_to_use = points.shape[1] * 2
         self.neighbours_to_use = neighbours_to_use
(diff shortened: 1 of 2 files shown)


plt109 (changes requested):
Nice improvement to `InterpolateAndExtrapolate` with the explicit typecasting of the maps to be interpolated. This should help ensure that maps are loaded properly even if they contain strings like 'null'.

Usually, one can create a custom class that inherits from `SR1Source` and overwrite the relevant functions, like `s2_acceptance` in this case. 

However, you showed that it is important to include this cut of s2>200 PE because its impact on the cs2 distribution. Could you also attach the plots you showed me yesterday that illustrated this? They will explain the motivation of the cut and will be good for documentation purposes.

CristianAntochi:
So, the point of this change is that without this acceptance cut the simulation will produce events below the minimum s1 of 200 pe, which is a hard selection cut on the data. 
This is particularly true for sources that are expected to produce low s2 i.e. wall model.  Here are some plots from the ongoing work on the wall source:

wall_s1_s2_no_acc

wall_cs1_cs2_no_acc

As you can see the source simulates events below the minimum s2 allowed.

PR #139, opened 2021-06-18
Documentation update: source attributes, block system, sphinx docstrings

JelleAalbers:
This is a documentation update:

  * Add comments with `#:` for attributes of `Source`, `BlockSource`, and a few other classes. These will appear in our [API reference](https://flamedisx.readthedocs.io/en/latest/reference/flamedisx.html#module-flamedisx.source) after this is merged.
  * Fix some problems with the block system documentation:
    * Several subsection headings did not appear;
    * The arguments to `_compute ` were never described;
    * Model attributes, which are relatively rare, were more prominently documented than model functions and dimensions, which are both used in every block. They were also occasionally referred to as 'static attributes', their old name.
  * Fix a few docstrings that weren't compatible with sphinx, and thus looked weird in the API reference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ docs/_build
 # Pycharm
 .idea
 
+# Visual studio code
+.vscode
+
 # coverage
 .coverage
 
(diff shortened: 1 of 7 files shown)


plt109 (approved):
Thanks Jelle! Nice to have a more comprehensive documentation.

JelleAalbers:
Thanks for reviewing Pueh!

PR #140, opened 2021-06-18
Configuration files and related improvements

JelleAalbers:
This makes it easier to change many settings of flamedisx sources in bulk. This is useful for saving and distributing fit results, or for quickly changing many detector-specific settings of a source.
  * You can now override model attributes, and model functions set to constants, in the same way you set parameter defaults: (through **kwargs to `Source.__init__`, `Source.set_data`, or `Source.set_params`, or `defaults` of `Likelihood.__init__`, see #103).
  * Adds a `config` argument for use in any of the aforementioned places. This should point to a python file, or name a file (without `.py`) from the new folder `flamedisx/configs`. Flamedisx will load the file and treat any variables in it as new defaults (for model attributes, parameters and constant model functions). An [example config file](https://github.com/FlamTeam/flamedisx/blob/64183217d86d560b6b44fceeb84d92a780c60b07/flamedisx/configs/example.py) is included. You can load multiple configs in the expected way, by giving `config` a list/tuple of (file)names .
  * Naming unused settings, either explicitly or via the config file, now triggers a warning. Previously you would hear no complaints if you did `fd.ERSource(hi_there=42)`, now you get `UserWarning: Defaults for unused options ignored: {'hi_there': 42}`. This is mostly useful for catching misspelled arguments.
  * Adds a `Source.print_config` method that will quickly give you an overview of all parameters and constant model functions, i.e. anything that you can override in a config file or via **kwargs/defaults (parameters, of course, can also be fitted). For example, `fd.ERSource().print_config()` gives:
```
Parameter                               Default             
------------------------------------------------------------
elife                                   452000.0            
er_pel_a                                15.0                
er_pel_b                                -27.7               
er_pel_c                                32.5                
er_pel_e0                               5.0                 
extraction_eff                          0.96                
g2                                      20.0                

Constant (could be made a function)     Default             
------------------------------------------------------------
double_pe_fraction                      0.219               
electron_acceptance                     1.0                 
electron_gain_std                       5.0                 
energy_spectrum_rate_multiplier         1.0                 
photoelectron_gain_mean                 1.0                 
photoelectron_gain_std                  0.5                 
photon_detection_eff                    0.1                 
work                                    0.0137              

Other attribute                         Default             
------------------------------------------------------------
check_acceptances                       True                
check_efficiencies                      True                
drift_velocity                          0.0001335           
energies                                tf.Tensor([ 0.   ...
fv_high                                 0.0                 
fv_low                                  -97.6               
fv_radius                               47.9                
rates_vs_energy                         tf.Tensor([1. 1. ...
t_start                                 2019-09-01 08:28:00 
t_stop                                  2020-09-01 08:28:00 
```
You can set the `format` argument of `print_config` to `config` to get the same information in a `variable = value` format. If you want to use this as a config file directly, you will also have to use e.g. `omit=('energies', 'rates_vs_energy', 't_start', 't_stop')'` to omit attributes like `energies` with defaults that do not `str()` to legal python literals. (And to fully automate config file generation, you'll have to [redirect stdout](https://stackoverflow.com/a/4675744) to capture the prints. If we end up using this we can of course simply this process.)

This is complementary to subclassing as a way of setting model functions, e.g. this
```python
class MySource(fd.ERSource):

    # A copy of ERSource.electron_detection_eff with the default elife changed
    @staticmethod
    def electron_detection_eff(drift_time, *, elife=987654, extraction_eff=0.96):
        return extraction_eff * tf.exp(-drift_time / elife)

    photon_detection_eff = 0.11

s = MySource()
```
accomplishes the same thing as
```python
s = fd.ERSource(config='example')
```
or
```python
s = fd.ERSource(elife=987654, photon_detection_eff=0.11)
```
However, subclassing is the most flexible method, since it also allows you to specify completely new model functions. You can of course use config files on your subclassed sources; the order of precedence is: explicitly given kwargs > config files > class attributes (for model functions) and hardcoded defaults (for parameters), i.e. the reverse of the examples above.

This is inspired by discussions with Robert following https://github.com/robertsjames/flamedisx/pull/1; I hope this will be able to address some of the issues discussed there.

Probably this will need a page in the documentation or a discussion in the tutorial.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  ignore:
+  - dependency-name: numpy
+    versions:
+    - 1.20.1
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ resource_cache
 # Data
 *.pkl
 *.csv
+
+# Apple
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ docs/_build
 # Pycharm
 .idea
 
+# Visual studio code
+.vscode
+
 # coverage
 .coverage
 
--- a/docs/source/blocks.rst
+++ b/docs/source/blocks.rst
@@ -52,7 +52,14 @@ Flamedisx's sources are built from units called **blocks**. Each block takes car
 
 Each block is represented by a :py:class:`~flamedisx.block_source.Block` class, which has three main methods:
 
-* :py:meth:`~flamedisx.block_source.Block._compute` is a block's main method. This returns a factor in the deterministic differential rate computation; for example, the probability of seeing the observed S2 for a range of possible number of detected electrons.
+* :py:meth:`~flamedisx.block_source.Block._compute` is a block's main method. This returns a factor in the deterministic differential rate computation; for example, the probability of seeing the observed S2 for a range of possible number of detected electrons. :py:meth:`~flamedisx.block_source.Block._compute` gets several arguments:
+
+  * `data_tensor` and `ptensor`: don't worry about these, just pass them to `gimme` when you call model functions -- see `Model functions`_ below.
+
+  * Keyword arguments with the (cross-)domains of the block -- see `Block dimensions`_ below.
+
+  * For blocks with dependencies, keyword arguments of the dependency's domain and result -- see `Block dependencies`_ below.
+
 * The :py:meth:`~flamedisx.block_source.Block._simulate` method performs a Monte Carlo simulation of the block's process. For example, drawing one possible S2 integral value, given a simulated event of a certain number of detected electrons.
 
   * Simulated events have a special `p_accepted` column, starting out at 1.0 for each event. Blocks can multiply this with probabilities for passing various selections. At the end of the simulation, a random number will be drawn to determine whether each event actually passes the selections.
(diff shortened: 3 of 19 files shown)


plt109:
Really neat new feature to have!

It will not only make things easier for adapting to new detector configurations and for parallelizing jobs on the cluster but also to keep track of configurations used for a particular fit/batch of MC generation.

However, does this only handle the tracking and management of single-valued numbers like the parameters in the model functions listed by `source().f_params`? How would/should one set the paths to the various maps (elife maps, light yield maps, etc) such as in https://github.com/FlamTeam/flamedisx/blob/f6ded37508d89b596217255ec504f1783300935a/flamedisx/xenon/x1t_sr1.py#L41 ?

JelleAalbers:
Thanks Pueh! That's a good point. I updated this branch so that any `model_attributes` (such as `fv_radius` and `t_start`) are now also configurable.

To make your maps configurable, you could try something like this (untested):
  
```python
class SR1Source:
    model_attributes = ('path_reconstruction_bias_mean_s1', 'other_map_path', 'etc')

    path_reconstruction_bias_mean_s1 = ('ReconstructionS1BiasMeanLowers_SR1_v2.json', 'ReconstructionS1BiasMeanUppers_SR1_v2.json')
    #... same for other maps
    
    def set_defaults(*args, **kwargs):
        super().set_defaults(*args, **kwargs)
        
        self.recon_map_s1_tf, self.domain_def_s1 = \
            read_maps_tf(path_reconstruction_bias_mean_s1, is_bbf=True)
        #... same for other maps
```

robertsjames:
I think this is a very nice step towards a config. system that general users will be able to use without diving into FD classes too deeply. For now with the NEST sources I will keep the current config. system I have there, and transition over to this one once that work is all done. I see some merge conflicts here however; @JelleAalbers are you happy to fix these (they look simple enough that they can be done even in the web editor!).

JelleAalbers:
Thanks Robert! Indeed, the merge conflicts were simple. I agree it makes sense to hold off on deconstructing your config system until you reach (or are at least closer to) the bottom of the stack of unmerged NEST work.

@plt109 with your SR1 modifications in, are you also OK with merging this, or would you like to see some other changes first?

plt109:
Thank Jelle! Please proceed with the merging. Eggcited to use this tmr!

PR #143, opened 2021-06-21
Wall source

CristianAntochi:
This update provides a source for surface events in xenon TPCs. The source is based on an ER source with spatial dependence.
To function the source needs a spatial template to be given. To create the spatial template a utility function is also provided: 
The function `construct_exponential_r_spatial_hist` can be used to construct a 3D multihist to be given to the wall source.

The wall source is constructed from a simplified flat ER band, with a much reduced electron extraction efficiency to account for the charge loss inside the TPC. The parameters of the source were tuned through fitting wall leakage data from XENON1T. 

The source and the fitting describes well enough the wall events and provides an alternative with some physical interpretation. The source can be reduced to have only three fittable parameters: `w_er_pel_e0`, `w_extraction_eff` and `w_q2` to account for any upwards shifting and spread in S2. In principle with an abundance of data, all parameters could be fitted.

Here are some plots of the comparison 2d histogram constructed with 20M wall simulated events from Flamedisx and wall data from XENON1T:
S2 vs R
wall_r_s2

Z vs R
wall_r_z-2

S1-S2 distribution (colorbar in log scale)
wall_s1_s2-2

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -320,7 +320,7 @@ class SR1Source:
                       cs1,
                       # Only used here, DEFAULT_.. would be super verbose
                       cs1_min=3.,
-                      cs1_max=150.):
+                      cs1_max=70.):
         acceptance = tf.where((cs1 > cs1_min) & (cs1 < cs1_max),
                               tf.ones_like(s1, dtype=fd.float_type()),
                               tf.zeros_like(s1, dtype=fd.float_type()))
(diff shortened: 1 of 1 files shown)


plt109:
Hello Cristian, thanks for the really nice work. Just to summarize what you explained to me verbally in person:

To use this source, the user would have to create the 3D multihist histogram (template) that describes how the wall events are distributed as a function of z, r, and theta using actual wall events using the function `construct_exponential_r_spatial_hist`. The user then passes this 3D template into the wall source and perform a fit with the same wall events to obtain an estimate for the shape parameters of `p_electron`, namely: `w_er_pel_e0`, `w_extraction_eff` and `w_q2`.

To illustrate the performance of this source, you fitted the wall events from 1T SR1. Following the procedure described above, you obtained a wall model, and then drew 20e6 events from the model and binned those events in a 2d histogram in s2 against r^2, z against r and s2 against s1, which are shown in colour in the plots. The actual events used for the fits are shown in red.
Only wall events up to 150 PE in S1 were used because the events are too sparse for a good fit. This range in S1 is to be determined separately for nT wall data.

PR #147, opened 2021-06-28
Test config

plt109:
As a follow-up test suggested by @JelleAalbers in https://github.com/FlamTeam/flamedisx/pull/140#issuecomment-864475656, the main changes in this PR occur in `x1t_sr1.py` the paths of the elife, cut efficiency, reconstruction bias mean, etc maps are moved inside `class SR1Source` so that the map paths are configurable as well.

The change in `itp_map.py` insists of specific type casting of the map values so that the new nT maps could be read in as well.

The changes are then tested on a local machine with 
```
class newSource(fd.SR1ERSource):
       # overwriting some methods

specificSource = newSource(config=my_config_path)
specificSource.print_config(column_widths=(40, 99))
```
returning the output shown in the screenshot, which displayed the expected values.
Screenshot from 2021-06-28 19-06-14

### DIFF
--- a/flamedisx/xenon/itp_map.py
+++ b/flamedisx/xenon/itp_map.py
@@ -28,7 +28,7 @@ class InterpolateAndExtrapolate:
         averaging. Default is 2 * dimensions of points.
         """
         self.kdtree = cKDTree(points)
-        self.values = values
+        self.values = values.astype(np.float)
         if neighbours_to_use is None:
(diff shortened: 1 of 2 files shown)


JelleAalbers (approved):
Looks good Pueh! Thanks for testing / implementing the new config system so quickly. I just removed some unused things (import json, __init__ that just called its super) and whitespace.

PR #152, opened 2021-07-14
Various bug fixes for NEST sources.

robertsjames:
Some small bug fixes for the NEST sources:

- Fix syntax in LUX sources after Jelle's changes
- Avoid S1 photoelectron detection binomial returning `nan`
- Temporary fix to `_annotate()` in pe_detection.py: will return to properly fix this later
- Fix where one block was accidentally using TensorFlow for `simulate()`
- No stepping for the sharply peaked S1 photoelectron detection binomial

### DIFF
--- a/flamedisx/lux/lux.py
+++ b/flamedisx/lux/lux.py
@@ -7,6 +7,7 @@ import configparser
 import os
 
 import flamedisx as fd
+from .. import nest as fd_nest
 
 export, __all__ = fd.exporter()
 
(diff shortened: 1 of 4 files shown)


JelleAalbers:
Thanks Robert!
 - Sorry, I missed the fd->fd_nest change in the LUX source.
 - Sounds good. In the classic flamedisx sources we have a similar but probably equivalent scheme that should avoid NaNs from even being made, see this compute function: https://github.com/FlamTeam/flamedisx/blob/master/flamedisx/lxe_blocks/double_pe.py#L20.
 - OK, not sure what the problem is but maybe you'll describe it when you fix it :-). I guess this preserves the NEST matching because the bounds are wide enough to accomodate annotate being somewhat off? Can you remove or comment the p_det_mle assignment to prevent the lint error?
 - Nice! Hope this will fix the failure in https://github.com/FlamTeam/flamedisx/pull/150
 - Sure. If it is sharply peaked, idealy the bounds estimate would be tight enough that stepping never becomes necessary. But maybe your MC bounds will achieve that, and anyway this late dimension should not be a big issue to keep fully in memory.
 
So good to merge once the build succeeds.

PR #168, opened 2021-09-17
bug fix in add_extra_columns to only add cs1, cs2 for simulated events

plt109:
This pull request fixes a bug in `add_extra_columns` and updates the conditions under which `cs1` and `cs2` are computed so that they are only computed for the simulated events and not replaced in actual data.

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -277,13 +277,11 @@ class SR1Source:
         # Add cS1 and cS2 following XENON conventions.
         # Skip this if s1/s2 are not known, since we're simulating
         # TODO: This is a kludge...
-        if 's1' in d.columns:
+        if ('s1' in d.columns) and ('cs1' not in d.columns):
             d['cs1'] = d['s1'] / d['s1_relative_ly']
-        if 's2' in d.columns:
-            d['cs2'] = (
-                d['s2']
-                / d['s2_relative_ly']
-                * np.exp(d['drift_time'] / d['elife']))
+        if ('s2' in d.columns) and ('cs2' not in d.columns):
+            d['cs2'] = (d['s2'] / d['s2_relative_ly']
+                      * np.exp(d['drift_time'] / d['elife']))
 
 
     @staticmethod


CristianAntochi (commented):
Good catch (me)! We should merge ASAP!

JelleAalbers:
I agree this is confusing, but I think this was done on purpose (and not documented/explained). Sorry if I'm mistaken, it's been a while since I looked at this.

If cs1 as computed by flamedisx differs from the cs1 in the data, flamedisx's model uses different corrections than the data processing. The ideal solution is to fix flamedisx's model so there is agreement, but if that's impossible, I think you should throw out the data's cs1/cs2 and take flamedisx's corrections. 

If, instead, you use the data's corrections, flamedisx's simulate and compute become inconsistent. Then flamedisx is no longer computing the actual likelihood -- or even the likelihood of any real model. It's mu will derive from simulate, using 'flamedisx corrections', and its differential rate from something that uses 'real data corrections'. 

Maybe the cs1/cs2 differences are too small to affect the mu much, but large enough to affect the fit quality. Even if that's true, taking the data's cS1/cS2 will cause unexpected surprises down the line. You might get flamedisx to say it has a good fit, but if you then simulate events from your best-fit model (e.g. to draw density contours for a plot), they will not correspond to the model you actually fitted, since you made simulate use different corrections than compute.

CristianAntochi:
We checked and the differences can be as high as 20% for some events and the mean difference is somewhere around 5% which is quite big. We are investigating what could go wrong with the corrections and why this happens. We found this bug looking at goodness of fits that show that s1-s2 fits are close to perfect while cs1-cs2 are terrible with  big inconsistencies and mismodelling between MC and MC-fitted.
Best guess for now is the bad computation of cs1 (biggest effect) in flamedisx or an underestimation of g1. Other causes could be a bad interpolation of the yield maps, but it probably shouldn't result in more than 1% difference.

JelleAalbers:
Interesting Cristian! Hope you find the souce of the discrepancy. As I wrote above though, I think discarding the data's cs1/cs2 is not a bug, and this is not a fix but something that will cause more headaches down the line. Your simulation is still using flamedisx's corrections, but your fit is now using the data's corrections in part (for the diff rate) and flamedisx's corrections in part (for mu).

Feel free to ignore me of course, this only affects the XENON sources :-)

CristianAntochi:
Then probably we should revert the changes and keep the branch for further exploration.

PR #171, opened 2021-09-23
Rj bounds 2

robertsjames:
Summary [here](https://docs.google.com/presentation/d/1Fr-_AQB__oQiLIyqbk3XxSAIWpivbzK7OnKqnoOWspM/edit#slide=id.p)

This PR overhauls the bounds estimation system. It is done in a more generalised way, via Bayes' theorem, and has been coded up in an automated way for binomial and normal blocks, which is all that is currently needed for the default and NEST sources. No more approximations or mystery equations ;)

This means that fluctuations are fully captured at all stages in the bounds computation, max_sigma no longer needs to be artificially inflated (3 is sufficient for both sets of sources up to 200 keV), and max_dim_size for the variable stepping can be lowered (120 -> 70, for both sets of sources).

I spent an awfully long time originally trying to accomplish this in a different way, via MC simulation. In hindsight this was dumb, but I've documented it [here](https://docs.google.com/presentation/d/155uRElYxMTmYaw55H7tzY_tOtteW5Efj5diaFmdHr5s/edit#slide=id.p), for posterity.


CristianAntochi:
Please always wait for reviews before merging PRs in the master!

PR #174, opened 2021-09-27
Rj bounds 2

robertsjames:
From #171:

---

Summary [here](https://docs.google.com/presentation/d/1Fr-_AQB__oQiLIyqbk3XxSAIWpivbzK7OnKqnoOWspM/edit#slide=id.p)

This PR overhauls the bounds estimation system. It is done in a more generalised way, via Bayes' theorem, and has been coded up in an automated way for binomial and normal blocks, which is all that is currently needed for the default and NEST sources. No more approximations or mystery equations ;)

This means that fluctuations are fully captured at all stages in the bounds computation, max_sigma no longer needs to be artificially inflated (3 is sufficient for both sets of sources up to 200 keV), and max_dim_size for the variable stepping can be lowered (120 -> 70, for both sets of sources).

I spent an awfully long time originally trying to accomplish this in a different way, via MC simulation. In hindsight this was dumb, but I've documented it here, for posterity.

---

Update:

As requested, only NEST sources now use Bayes bounds. Default sources bounds are untouched.


flamedisx/block_source.py
```diff
@@ -144,6 +147,117 @@ def annotate(self, d: pd.DataFrame):
                           <= d[f'{dim}_max'].values), \
                 f"_annotate of {self} set misordered bounds"
 
+    def bayes_bounds_binomial(self, df, in_dim, supports, rvs_binom, ns_binom, ps_binom, bound):
```
JelleAalbers, line 150: I think the bayes_bounds_* functionswould be best as standalone functions in a separate file, rather than among the infrastructure in block_source.py. If you give bounds_prob as an extra argument, the functions won't depend on `self` anymore. Apart from decluttering block_source.py, this will make it easier to unit test the bayes bound functions, and compare them directly to the new implementations in terms of compute later.

flamedisx/block_source.py
```diff
@@ -144,6 +147,117 @@ def annotate(self, d: pd.DataFrame):
...
+        must be the same shape as supports
+        :param ps_binom: Variable the block uses as the success probability of the binomial calculation;
+        must be the same shape as supports
+        :param bound: 'upper', 'lower' or 'mle' to determine which bound is currently being calculated
+        """
+        assert (bound == 'upper' or 'lower' or 'mle'), "bound argumment must be upper, lower or mle"
+        assert (np.shape(rvs_binom) == np.shape(ns_binom) == np.shape(ps_binom) == np.shape(supports)), \
+            "Shapes of suports, rvs_binom, ns_binom and ps_binom must be equal"
+
+        def binomial(x, n, p):
```
JelleAalbers, line 170: Why not use scipy.stats' functions for this and other basic distribution calculations below? E.g. `stats.binom(n=.., p=..).pmf(k)`. These are fast, vectorized, extensively tested, and they will even use appropriate approximations when possible. Flamedisx also depends on scipy already.

flamedisx/block_source.py
```diff
@@ -459,6 +587,50 @@ def _annotate(self, _skip_bounds_computation=False):
         for b in self.model_blocks[::-1]:
             b.annotate(d)
 
+    def scale_observables(self, df, observables):
```
JelleAalbers, line 590: The `scale_observables` and `energy_bounds` methods are currently unused (if I understood correctly) and undocumented. I understand you'll want to propose them in the next PR, happy to take a look at them then -- in the meantime, could you move them somewhere else or remove them?

For the KDE, any reason not to use scipy's KDE? I think it is [faster](https://kdepy.readthedocs.io/en/latest/comparison.html) and would avoid adding scikit-learn as a dependency to flamedisx.

flamedisx/source.py
```diff
@@ -29,7 +31,10 @@ class Source:
     #: Whether to trace (compile into a tensorflow graph) the differential
     #: rate computation
     trace_difrate = True
+
     default_max_sigma = 3
+    default_max_dim_size = 70
```
JelleAalbers, line 36: Could you put the default back to 120, now that the default bounds are unchanged? Or did you test that 120 is overkill even with the old bounds?

flamedisx/source.py
```diff
@@ -47,6 +52,10 @@ class Source:
     #: for use in domain / cross-domain
     final_dimensions: ty.Tuple[str] = tuple()
 
+    #: Names of penulitmate model dimensions;
```
JelleAalbers, line 55: OK, though in the future we could autodetect these, or setup a more general system of specifying default max dimsizes per block.

flamedisx/source.py
```diff
@@ -205,8 +217,17 @@ def __init__(self,
         """
         if max_sigma is None:
             max_sigma = self.default_max_sigma
+        self.bounds_prob = 0.5 * (1 + sp.erf(-max_sigma / np.sqrt(2)))
```
JelleAalbers, line 220: `stats.norm.cdf(-self.max_sigma)` would be clearer (with `from scipy import stats` in the imports of course)

flamedisx/source.py
```diff
@@ -276,6 +297,8 @@ def __init__(self,
         if not _skip_tf_init:
             self.trace_differential_rate()
 
+        self.source_copy = deepcopy(self)
```
JelleAalbers, line 300: I think we don't need this yet, let's have a look at this in your next PR -- maybe there is another way

flamedisx/source.py
```diff
@@ -742,6 +768,9 @@ def mu_before_efficiencies(self, **params):
         """
         raise NotImplementedError
 
+    def setup_copy():
```
JelleAalbers, line 771: same here

flamedisx/lxe_blocks/energy_spectrum.py
```diff
@@ -120,6 +119,9 @@ def validate_fix_truth(self, d):
             # Assume fix_truth is a one-line dataframe with at least
             # cols columns
             return d[cols].iloc[0].to_dict()
+        elif isinstance(d, pd.Series):
+            cols = ['x', 'y', 'z', 'r', 'theta', 'event_time', 'drift_time']
```
JelleAalbers, line 123: Duplicates line 118; you can define it outside the if

flamedisx/lxe_blocks/quanta_generation.py
```diff
@@ -149,8 +149,8 @@ def _compute(self,
         work = self.gimme('work', data_tensor=data_tensor, ptensor=ptensor)
         mean_q_produced = (
                 energy_noStep
-                * self.gimme('lindhard_l', bonus_arg=energy_noStep,
-                             data_tensor=data_tensor, ptensor=ptensor)
+                * self.gimme('lindhard_l', bonus_arg=energy_noStep[:, 0, :],
```
JelleAalbers, line 152: Can you explain what is happening here? What are the explicit [:, 0, :]'s doing / why do we need them now and not before?

JelleAalbers (changes requested):
Thanks Robert! The Bayes bounds are a key improvement for flamedisx that we'll want to have for all sources soon. This supersedes most of the heuristics currently in annotate, and should make all of flamedisx faster (when the bounds can be tight) or more reliable (when they can't).

As discussed on slack, the current implementation essentially duplicates the implementation of `compute`, which is not very maintainable. I'll open an issue on this. For the moment, since it only affects the NEST sources and you want to move on to other features, I'm happy to go ahead regardless.

Could you have a look at the comments I left on the code? Once you reset the default, I don't think this PR should cause any changes in the main sources. @plt109 or @CristianAntochi would you like some time to test that yourself, or do you agree?

robertsjames:
Thanks for the comments Jelle - addressing them now. I'd strongly advocate that we **don't** squash merge this - as I'm tentatively removing things which I may need to come back and add later, I'd like a copy of my thought process to be preserved. Having the full history will also make tracing potential yet-to-be-spotted bugs easier. And then on a slightly more selfish note, it means my git activity will properly reflect the level of work I did, which is always a nice thing

flamedisx/block_source.py
robertsjames: Done

flamedisx/source.py
robertsjames: Agree

flamedisx/source.py
robertsjames: Removed for now. I'd advocate that we please don't squash-merge this: I'd like a copy of my thought process to be preserved, if this comes back to bite in future

flamedisx/source.py
robertsjames: Removed

JelleAalbers:
Sure, we can try to rebase-merge then; this would mean all your commits are preserved but follow directly after each other (though they retain their original timestamps). A regular merge would scatter your commits all over the history.

flamedisx/block_source.py
robertsjames: Done; it was a k-d tree, not KDE, but yeah, I'm using the SciPy version now. I moved the `energy_bounds` stuff to bounds.py also; hopefully this is fine

flamedisx/lxe_blocks/quanta_generation.py
robertsjames: This was just to speed things up: before, `lindhard_l` was evaluated over a tensor with dimensions (energy_spectrum_length, nq_max - nq_min, events). But the function only depends on energy, not nq. So the computation was being repeated unnecessarily, and it was quite time consuming too, which meant the default NR models scaled quite poorly for longer energy spectra. This evaluates the function just over the dimensions (energy_spectrum_length, events), and then repeats it over nq.

flamedisx/block_source.py
robertsjames: Done, tested, all good

flamedisx/source.py
robertsjames: 70 was fine for the default sources, even with the old bounds

flamedisx/lxe_blocks/energy_spectrum.py
robertsjames: Done

flamedisx/lxe_blocks/quanta_generation.py
JelleAalbers: Thanks, that's a good optimization!

flamedisx/source.py
JelleAalbers: That's good to hear, but it might affect the tests Pueh and Cristian are currently doing, and you wrote to them in the slack that the default sources would be unaffected. Since you are eager to merge quickly, can you leave the current default and open a separate PR for them to look at? @plt109 and @CristianAntochi feel free to chime in if I'm misunderstanding.

flamedisx/block_source.py
JelleAalbers: Great, the stuff in bounds.py looks a lot cleaner/clearer now.

flamedisx/source.py
JelleAalbers: Resolved via slack

JelleAalbers (approved):
Thanks for the changes Robert, all good to go!

PR #176, opened 2021-09-27
Fix hessian bug

plt109:
This PR fixes a bug in the computation of the covariance matrix used in `LogLikelihood.summary` and the computing a starting guess for the interval computation.

Mathematically, the estimator of the asymptotic covariance matrix is the inverse of the Fisher information matrix evaluated at the bestfit parameters of the fit. See [stack exchange article here](https://stats.stackexchange.com/questions/68080/basic-question-about-fisher-information-matrix-and-relationship-to-hessian-and-s) (or pg 101-102 of Statistical Methods in Experimental Physics 2nd Edition by Fred James).

`LogLikelihood.inverse_hessian` returns the inverse of the hessian of -2 loglikelihood, whereas the covariance is the inverse of the hessian of -loglikelihood.

The implementation of `LogLikelihood.inverse_hessian` is kept unchanged because conceptually, the object being minimised is indeed the -2 loglikelihood, and for cleanliness of the code `LogLikelihood.inverse_hessian` should return the inverse hessian of the object being minimised, which is the -2 loglikelihood.

The correct inverse hessian (inverse hessian of the -loglikelihood) is then passed into `cov_to_std`. The keyword argument `inverse_hessian` of `LogLikelihood.summary` is also changed to `cov` for clarity (that this is the inverse hessian of the -loglikelihood and not the inverse hessian of the -2 loglikelihood).

The following code snippet from @JelleAalbers is used to test the changes in the PR.
```
import numpy as np
import pandas as pd

import flamedisx as fd

class MySource(fd.ColumnSource):
    column = 'constant_diff_rate'
    mu = 1

n_obs = 10_000
data = pd.DataFrame(dict(constant_diff_rate=np.ones(n_obs)))

ll = fd.LogLikelihood(
    sources=dict(source=MySource), 
    data=data,
    batch_size=10_000,
    free_rates='source')

# This should be close to n_obs
bf = ll.bestfit(guess=dict(source_rate_multiplier=1.))
print(bf['source_rate_multiplier'], n_obs)

inv_hess = ll.inverse_hessian(bf)

ll.summary(
    bf, 
)
```

A source with constant diff rate and 10k observed events should give a 1 sigma error of +- 100. 

Using the current `main` branch (97d3862c0691e1f8e07efc220dea3cf03c646f3c), the computed 1 sigma error is incorrect (see screenshot below)
wrong_std

Using this `fix_hessian_bug` branch(5a32d66bd987f13c789d8f27e655a5dbbab1087c), the computed 1 sigma error is as expected (see screenshot below
correct_std
)

### DIFF
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -626,8 +626,11 @@ class IntervalObjective(Objective):
 
         if sigma_guess is None:
             # Estimate one sigma interval using parabolic approx.
+            # inverse_hessian returns hessian of -2 loglikelihood but the
+            # covariance matrix is calculated from the hessian of -
+            # loglikelihood
             sigma_guess = fd.cov_to_std(
-                self.lf.inverse_hessian(bestfit)
+                2. * self.lf.inverse_hessian(bestfit)
             )[0][self.arg_names.index(self.target_parameter)]
         self.sigma_guess = sigma_guess
 
--- a/flamedisx/inference.py
+++ b/flamedisx/inference.py
@@ -626,11 +626,11 @@ class IntervalObjective(Objective):
 
         if sigma_guess is None:
             # Estimate one sigma interval using parabolic approx.
-            # inverse_hessian returns hessian of -2 loglikelihood but the
-            # covariance matrix is calculated from the hessian of the
-            # - loglikelihood
+            # `inverse_hessian` returns inverse of the hessian of -2 loglikelihood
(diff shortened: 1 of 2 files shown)


JelleAalbers (approved):
Nice catch Pueh, thank you! I like the name changes in `summary` too.

CristianAntochi:
Uh! this is the thing we talked about. It is indeed a good catch!

PR #178, opened 2021-10-05
Simplified p_electron and less params for the wall

CristianAntochi:
This pull request makes the wall source simpler:
`p_electron` has now only one parameter and it is modelled as constant.
`electron_gain_mean` and `electron_gain_std` depends on `w_g2_total`, a fittable parameter, and `w_extraction_eff`

The variation of the parameters `w_q2`, `w_g2_total` and `w_er_pel_c` are enough to account for most of the possible wall event distributions in S1-S2. 

One possible further simplification would be to make also the `p_electron_fluctuation` just a single parameter.

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -305,10 +305,16 @@ class SR1Source:
         # 0 * light yield is to fix the shape
         return single_electron_width + 0. * s2_relative_ly
 
+    @staticmethod
+    def double_pe_fraction(z, *, dpe=DEFAULT_P_DPE):
+        # Ties the double_pe_fraction model function to the dpe
+        # parameter in the sources
+        return dpe + 0 * z
+
     #TODO: implement better the double_pe_fraction or photon_detection_efficiency as parameter
     @staticmethod
-    def photon_detection_eff(s1_relative_ly, g1=DEFAULT_G1):
-        mean_eff= g1 / (1. + DEFAULT_P_DPE)
+    def photon_detection_eff(s1_relative_ly, g1=DEFAULT_G1, dpe=DEFAULT_P_DPE):
+        mean_eff= g1 / (1. + dpe)
         return mean_eff * s1_relative_ly
 
     def photon_acceptance(self,
@@ -362,15 +368,15 @@ class SR1Source:
 class SR1ERSource(SR1Source, fd.ERSource):
 
     @staticmethod
-    def p_electron(nq, *, W=13.8e-3, mean_nexni=0.15,  q0=1.13, q1=0.47,
-                   gamma_er=0.031 , omega_er=31.):
+    def p_electron(nq, *, W=13.7e-3, mean_nexni=0.15,  q0=1.13, q1=0.47,
+                   gamma_er=0.031 , omega_er=31., delta_er=0.24):
         # gamma_er from paper 0.124/4
         F = tf.constant(DEFAULT_DRIFT_FIELD, dtype=fd.float_type())
 
         e_kev = nq * W
         fi = 1. / (1. + mean_nexni)
         ni, nex = nq * fi, nq * (1. - fi)
-        wiggle_er = gamma_er * tf.exp(-e_kev / omega_er) * F ** (-0.24)
+        wiggle_er = gamma_er * tf.exp(-e_kev / omega_er) * F ** (-1.*delta_er)
 
         # delta_er and gamma_er are highly correlated
         # F **(-delta_er) set to constant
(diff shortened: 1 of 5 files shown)


flamedisx/xenon/x1t_sr1.py
```diff
@@ -439,40 +443,62 @@ class SR1WallSource(fd.SpatialRateERSource, SR1ERSource):
...
-          # it has to be a direct function of nq.
-          e_kev_sortof = nq * 13.7e-3
-          eps = fd.tf_log10(e_kev_sortof / w_er_pel_e0 + 1e-9)
-          qy = (
-              w_er_pel_a * eps ** 2
-              + w_er_pel_b * eps
-              + w_er_pel_c)
-          return fd.safe_p(qy * 13.7e-3)
+          # TODO: Maybe this can be tied to W in a combined fit
+          W = 13.8e-3 #keV
```
plt109, line 452: The value to use should be the literature value, 13.7 eV. This is the posterior mean from 1T BBF fits.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -439,40 +443,62 @@ class SR1WallSource(fd.SpatialRateERSource, SR1ERSource):
...
-            float('inf'))
-      # It is preferred to have higher energy spectrum for the wall
-      energies = tf.cast(tf.linspace(0., 25. , 1000),
-                       dtype=fd.float_type())
+      def electron_loss(electrons_produced,
+                        *,
+                        l_max=0.084,
+                        alpha=0.0001,
+                        l_min=0.002):
+          return l_max*(tf.exp(-alpha*electrons_produced))+l_min
```
plt109, line 470: As discussed, since `electron_loss` is a value between [0,1], there should be a `safe_p` wrapping this, or a `tf.clip_by_value` if a typical value is constrained to an even smaller range than [0,1].

flamedisx/xenon/x1t_sr1.py
```diff
@@ -439,40 +443,62 @@ class SR1WallSource(fd.SpatialRateERSource, SR1ERSource):
...
+          return g2 * tf.ones_like(r)
+
+      @staticmethod
+      def electron_gain_std(r,
+                            *,
+                            w_g2=25.5,
+                            w_extraction_eff=0.89):
+          # 0 * r is to fix the shape
+          single_electron_width=w_g2/w_extraction_eff*0.25
+          return single_electron_width + 0. * tf.ones_like(r)
```
plt109, line 485: `0. * r` instead of `0. * tf.ones_like(r)`.

plt109 (changes requested):
This class would be very important for the user to construct a data-driven surface background model which would be part of any fits in a WIMP search. The configuration of all sources used in a combined fit are currently set with a single config file via the config file functionality added in [PR 140](https://github.com/FlamTeam/flamedisx/pull/140). It would be very important for this surface background source class to be compatible with the config file functionality.

Screenshot from 2022-03-22 11-36-43

flamedisx/xenon/x1t_sr1.py
CristianAntochi: If I remember correctly * r only gave problems with tensorflow

CristianAntochi:
> This class would be very important for the user to construct a data-driven surface background model which would be part of any fits in a WIMP search. The configuration of all sources used in a combined fit are currently set with a single config file via the config file functionality added in [PR 140](https://github.com/FlamTeam/flamedisx/pull/140). It would be very important for this surface background source class to be compatible with the config file functionality.
> 
> Screenshot from 2022-03-22 11-36-43

What happens here is that you call the source without a spatial histogram, and the source is not properly initialised. To run a proper wall source you should build a spatial histogram outside of the source, possibly using the example of the utility function defined [here](https://github.com/FlamTeam/flamedisx/blob/a76b2d58b1fbd38bcd0ce28a0796fd12e61b9861/flamedisx/xenon/x1t_sr1.py#L119), then define your own wall source as
```
def MyWall(SR1WallSource):
      spatial_hist = normalised_Histdd_wall_spatial_hist
```
It would be beneficial if fiducial volume and additional acceptances are defined in the source as well, to bypass possible acceptance cuts inherited from other sources.

flamedisx/xenon/x1t_sr1.py
CristianAntochi: solved

flamedisx/xenon/x1t_sr1.py
CristianAntochi: solved

flamedisx/xenon/x1t_sr1.py
CristianAntochi: solved

JelleAalbers:
No activity for two years -> let's close this. If you're reading this, thanks for your work Cristian! I think @cecilia-ferrari is working on a different wall model implementation, which will hopefully incorporate or build on some of the improvements you suggested here.

PR #193, opened 2022-01-17
Extraction efficiency detached

CristianAntochi:
This PR detaches the extraction efficiency from the electron detection efficiency in a similar fashion to the `penning_quenching` in the S1 branch of the detection block. 
This modification allows the `extraction_eff` to be a model function that can be non uniform throughout the detector and possibly become also a function of the number of `electrons_produced`. This allows for more flexible modelling of sources that have some special distribution of S2 (wall), or for non optimal detector conditions where the extraction efficiency is non uniform and depends on the number of produced electrons. 

The core functionality of FLAMEDISX is not impacted by this change and none of the base computation are broken and all the sources that do not use explicitly `extraction_eff` as a parameter are not affected. **Config files might need to change**.
This change breaks any source that uses `extraction_eff` as a parameter and not as a model function. Specifically it breaks the Xe1T sources. A very easy fix to those sources will be provided in the next PR.

### DIFF
--- a/flamedisx/lxe_blocks/detection.py
+++ b/flamedisx/lxe_blocks/detection.py
@@ -37,6 +37,10 @@ class DetectPhotonsOrElectrons(fd.Block):
             p = p * self.gimme('penning_quenching_eff',
                                bonus_arg=quanta_produced,
                                data_tensor=data_tensor, ptensor=ptensor)
+        else:
+            p = p * self.gimme('extraction_eff',
+                               bonus_arg=quanta_produced,
+                               data_tensor=data_tensor, ptensor=ptensor)
 
         result = tfp.distributions.Binomial(
                 total_count=quanta_produced,
@@ -53,6 +57,9 @@ class DetectPhotonsOrElectrons(fd.Block):
         if self.quanta_name == 'photon':
             p *= self.gimme_numpy(
                 'penning_quenching_eff', d['photons_produced'].values)
+        else:
+            p *= self.gimme_numpy(
+                'extraction_eff', d['electrons_produced'].values)
 
         d[self.quanta_name + 's_detected'] = stats.binom.rvs(
             n=d[self.quanta_name + 's_produced'],
@@ -67,6 +74,9 @@ class DetectPhotonsOrElectrons(fd.Block):
         if self.quanta_name == 'photon':
             eff *= self.gimme_numpy('penning_quenching_eff',
                                     d['photons_detected_mle'].values / eff)
+        else:
+            eff *= self.gimme_numpy('extraction_eff',
+                                    d['electrons_detected_mle'].values / eff)
 
         # Check for bad efficiencies
         if self.check_efficiencies and np.any(eff <= 0):
@@ -123,13 +133,17 @@ class DetectPhotons(DetectPhotonsOrElectrons):
 class DetectElectrons(DetectPhotonsOrElectrons):
     dimensions = ('electrons_produced', 'electrons_detected')
 
-    special_model_functions = ('electron_acceptance',)
+    special_model_functions = ('electron_acceptance', 'extraction_eff')
     model_functions = ('electron_detection_eff',) + special_model_functions
 
     @staticmethod
     def electron_detection_eff(drift_time, *,
-                               elife=452e3, extraction_eff=0.96):
-        return extraction_eff * tf.exp(-drift_time / elife)
+                               elife=452e3):
+        return 1. * tf.exp(-drift_time / elife)
+
+    @staticmethod
+    def extraction_eff(nel):
+        return 0.96 + 0. * nel
 
     electron_acceptance = 1.
 
--- a/tests/test_likelihood.py
+++ b/tests/test_likelihood.py
@@ -297,9 +297,8 @@ def test_hessian_rateonly(xes: fd.ERSource):
         """
         @staticmethod
         def electron_detection_eff(drift_time, *,
-                                   different_elife=333e3,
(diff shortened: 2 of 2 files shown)


flamedisx/lxe_blocks/detection.py
```diff
@@ -123,13 +133,17 @@ def _compute(self, data_tensor, ptensor,
 class DetectElectrons(DetectPhotonsOrElectrons):
     dimensions = ('electrons_produced', 'electrons_detected')
 
-    special_model_functions = ('electron_acceptance',)
+    special_model_functions = ('electron_acceptance', 'extraction_eff')
     model_functions = ('electron_detection_eff',) + special_model_functions
 
     @staticmethod
     def electron_detection_eff(drift_time, *,
```
JelleAalbers, line 140: Maybe you could rename electron_detection_eff to electron_drift_eff, to emphasize the extraction efficiency is no longer part of it?

flamedisx/lxe_blocks/detection.py
```diff
@@ -123,13 +133,17 @@ def _compute(self, data_tensor, ptensor,
...
-                               elife=452e3, extraction_eff=0.96):
-        return extraction_eff * tf.exp(-drift_time / elife)
+                               elife=452e3):
+        return 1. * tf.exp(-drift_time / elife)
+
+    @staticmethod
+    def extraction_eff(nel):
```
JelleAalbers, line 145: Could you make the extraction eff a keyword argument just like elife above / as it was before?

JelleAalbers (commented):
Good idea Cristian. Even for regular sources, the extraction eff model may be relatively complex (e.g. filling reconstructed positions into some spatially-dependent model function or map), so splitting extraction and drift efficiency into two separate model functions is sensible. 

Would be good to hear from Pueh and Robert about potential impacts on the XENON and NEST sources. If you have a quick fix for the XENON1T sources, maybe it's worth just adding that to this PR.

flamedisx/lxe_blocks/detection.py
CristianAntochi: Good point, fairly easy to add!

flamedisx/lxe_blocks/detection.py
```diff
@@ -123,13 +133,19 @@ def _compute(self, data_tensor, ptensor,
 class DetectElectrons(DetectPhotonsOrElectrons):
     dimensions = ('electrons_produced', 'electrons_detected')
 
-    special_model_functions = ('electron_acceptance',)
+    special_model_functions = ('electron_acceptance', 'extraction_eff')
```
JelleAalbers, line 136: Currently you have `extraction_eff` as a special model function, taking `electrons_produced` as a bonus argument. But electrons_produced is the number of electrons captured before impurities, the extraction efficiency shouldn't be able to see that. Could you make it a regular model function instead, like electrons_detected? Sorry, didn't spot this on the first pass.

flamedisx/lxe_blocks/detection.py
JelleAalbers: Ah you're right, since this class handles both photons and electrons that would indeed require extra if statements. Never mind.

flamedisx/lxe_blocks/detection.py
JelleAalbers: Thanks!

flamedisx/lxe_blocks/detection.py
CristianAntochi: Indeed it is a special model function at the moment and indeed in theory it shouldn't depend on electrons produced, however the point for the PR is that I wanted to make the `electron_detection_efficiency` a function of electrons produced somehow, and this is the easiest way I could think of. I could add another function to make it more physical, something like `electron_quenching` or `electron_loss` that depends on `electrons_produced` before the drift, and make the `extraction_eff` a regular model function with observable dependence and keyword arguments.

robertsjames:
All looks good to me - no impact at all on the NEST sources

robertsjames (approved):
All looks good to me - no impact at all on the NEST sources

JelleAalbers (approved):
Nice, I like the symmetry with penning quenching! We might even rename the penning quenching function to 'photon loss' at some point (though that would require some downstream changes). I just added one suggestion for a variable name.

flamedisx/lxe_blocks/detection.py
```diff
@@ -123,14 +133,20 @@ def _compute(self, data_tensor, ptensor,
...
     @staticmethod
     def electron_detection_eff(drift_time, *,
-                               elife=452e3, extraction_eff=0.96):
+                               extraction_eff=0.96,
+                               elife=452e3):
+        # TODO: change the function name, it is a drift_efficiency
         return extraction_eff * tf.exp(-drift_time / elife)
 
+    @staticmethod
+    def electron_loss(pel):
```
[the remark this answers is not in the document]
CristianAntochi: Makes sense! I will implement the change and should be good to go!

plt109 (approved):
Looks good to me, but would be good to have a comment to remind the user that `electron_loss` should be constrained between [0,1], and noting that numerical issues can occur when `electron_loss` is near 0 or 1.

CristianAntochi:
> Looks good to me, but would be good to have a comment to remind the user that `electron_loss` should be constrained between [0,1], and noting that numerical issues can occur when `electron_loss` is near 0 or 1.

That is true. We should probably specify in comments or doc-lines the sensible output for each model function, specifically the efficiencies that can be overwritten

PR #202, opened 2022-02-22
Add survival map to x1t_sr1 source

cecilia-ferrari:
## PR Description:
Added the possibility to include a survival map to better model the electron_detection_efficiency function

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -27,6 +27,8 @@ DEFAULT_DRIFT_VELOCITY = 1.34 * 1e-4   # cm/ns, from analysis paper II
 DEFAULT_DRIFT_FIELD = 81.
 DEFAULT_SURVIVAL_PROBABILITY = 1.
 
+DEFAULT_DRIFT_VELOCITY = 6.77*1e-5     # cm/ns, Valerio's note
+
 DEFAULT_G2_TOTAL = DEFAULT_G2 / (1.-DEFAULT_AREA_FRACTION_TOP)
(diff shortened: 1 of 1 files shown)


CristianAntochi:
Hello Cecilia! The PR looks good and should be very useful, however I think x1t_sr1.py should not have this modification since the maps refer to the nT maps while the sources are for 1T. I believe we should create a new nT source file for base nT sources that takes a private nT configuration file.
 I hope we can do that soon and that @plt109 agrees.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -356,9 +387,11 @@ def add_extra_columns(self, d):
     @staticmethod
     def electron_detection_eff(drift_time,
                                elife,
+                               survival_probability,
                                *,
                                extraction_eff=DEFAULT_EXTRACTION_EFFICIENCY):
-        return extraction_eff * tf.exp(-drift_time / elife)
+        # electron_detection_eff comprises both the elife and the CIV effects
+        return tf.clip_by_value(extraction_eff * tf.exp(-drift_time / elife) * survival_probability ,1e-18,1)
```
CristianAntochi, line 394: Why 1e-18 for the clipping? sounds a bit extreme and quite arbitrary

cecilia-ferrari:
I agree with you @CristianAntochi 
I can also help with this!

flamedisx/xenon/x1t_sr1.py
cecilia-ferrari: Because the map interpolation  returned sometimes very low values. This affected  the whole electron det eff and the algorithm didn't work.. 
Do you think e-18 is too low?

PR #205, opened 2022-03-10
Rj nest pre quanta

robertsjames:
Summary [here](https://docs.google.com/presentation/d/1CZJYkNbn1ozxGKHkDhzFp6O9agFfZl6wRM8M2bPHmRY/edit?usp=sharing)

This PR implements completes the NEST sources with the full yield models. Also includes updates to the Bayes bounds, and core changes needed for a successful and efficient NEST implementation. Default sources are _unaffected_


flamedisx/nest/lxe_blocks/quanta_splitting.py
```diff
@@ -2,110 +2,476 @@
...
+                nel_mean = self.gimme('mean_yield_electron', data_tensor=data_tensor, ptensor=ptensor,
+                                      bonus_arg=energy)
+                nq_mean = self.gimme('mean_yield_quanta', data_tensor=data_tensor, ptensor=ptensor,
+                                     bonus_arg=(energy, nel_mean))
+                fano = self.gimme('fano_factor', data_tensor=data_tensor, ptensor=ptensor,
+                                  bonus_arg=nq_mean)
+
+                p_nq = tfp.distributions.Normal(loc=nq_mean,
+                                                scale=tf.sqrt(nq_mean * fano) + 1e-10).cdf(nq + 0.5) \
+                    - tfp.distributions.Normal(loc=nq_mean,
```
JelleAalbers, line 66: Same distribution constructed twice here. The 'else' has the same issue twice

flamedisx/nest/lxe_blocks/quanta_splitting.py
```diff
@@ -2,110 +2,476 @@
...
+
+            r_final = p_final * rate_vs_energy
+
+            r_final = tf.where(tf.math.is_nan(r_final),
+                               tf.zeros_like(r_final, dtype=fd.float_type()),
+                               r_final)
+
+            return r_final
+
+        def compute_single_energy_approx(args):
```
JelleAalbers, line 133: This function looks very similar to `compute_single_energy_full`. Could they be one method with branching paths controlled by an argument?

flamedisx/source.py
```diff
@@ -598,14 +622,15 @@ def trace_differential_rate(self):
             self._differential_rate,
             input_signature=input_signature)
 
-    def differential_rate(self, data_tensor=None, autograph=True, **kwargs):
+    def differential_rate(self, i_batch=tf.constant(0, dtype=fd.int_type()),
```
JelleAalbers, line 625: Many Source and Block functions have now gained an extra `i_batch` argument, even high-level methods like `Source.differential_rate`. 

This is not very nice.. E.g. someone reading the code and seeing both data_tensor and i_batch as an argument would assume these methods take the full data tensor and slice a batch out of it. But that's not what happens. Changing so many functions also forced changes to many unrelated places.

Could you explain why you this was necessary, and did you consider and reject alternatives? I think you're using i_batch to access one or more new data-dependent tensors with bounds info -- if that's the only use, could you e.g. add the bounds info to a data_tensor column instead (and take it from the first event in a batch)? 

Creating new data-dependent tensors as attributes may also have other consequences. I can't recall the specifics (probably something related to re-initialization -- but I know we used to have data_tensor and ptensor as attributes, then decided to pass them around explicitly later. We wouldn't have done that without a good reason.

tests/test_inference.py
```diff
@@ -14,6 +14,7 @@ def test_one_parameter_interval_nonlincontr(xes):
 
     lf = fd.LogLikelihood(
         sources=dict(er=xes.__class__),
+        arguments=dict(er=dict()),
```
JelleAalbers, line 17: Can you init arguments to an empty dict by default? Then users won't have to add these dummy dicts all the time.

flamedisx/lxe_blocks/quanta_generation.py
```diff
@@ -15,7 +15,7 @@
 class MakeERQuanta(fd.Block):
 
     dimensions = ('quanta_produced', 'energy')
-    extra_dimensions = (('quanta_produced_noStep', False),)
+    bonus_dimensions = (('quanta_produced_noStep', False),)
```
JelleAalbers, line 18: Can you explain what is happening here? I.e. did you rename extra_dimensions to bonus_dimensions, does this relate to our earlier conversations about this attribute (in https://github.com/FlamTeam/flamedisx/pull/127) and the way the attributes are described in source.py (which autofeeds the docs)? Probably it's also related to returning an extra `quanta_produced` domain from `domain_dict_bonus`,

JelleAalbers (changes requested):
Just to be clear, I would suggest to address
  * the tensorflow-probability files;
  * the i_batch stuff, and
  * defaults for `Likelihood.__init__` `arguments` argument

before merging. And someone should sanity-check the new bounds stuff (I hope to look at it soon as well).

I think we can leave other things for later if that works better for the hackathon (you're working on other important issues too, and it's obvious we want this PR no matter what) but the above things change some pretty basic things about how people are going to install and use flamedisx.

robertsjames:
Thanks for taking a look so quickly Jelle! I will do my best to address all this tomorrow, so we can get it merged ASAP

flamedisx/source.py
robertsjames: The reason why I couldn't put these bounds in the regular data tensor is because for the ion bounds, we calculate a separate upper and lower bound for each energy in the trimmed/stepped energy spectrum being summed over (this massively improved the accuracy/speed tradeoff). As this results in a column of bounds whose entries are themselves a (1D) tensor, we can't stack this into the original data tensor with the right shape.

My first pass attempt at this (and something I just spent a while reconsidering just now) was to create this 'special tensor' of energy-dependent ion bounds and pass around batch slices of it in the same way as the data tensor. However this is problematic because a `source` does not know how many of these special tensors are going to be needed (they are created within a block), and which block of `block_source` will need to access each one...

So my conclusion was that the only sensible way is to pass around the batch number so the blocks can see it, and index into the current batch special tensors directly within the block's computation... I didn't see any obvious issues with passing them around as attributes, but I suppose I might have missed something. If you have any immediate alternative ideas I'm happy to try implementing them, but yes, this was somewhat of a last resort solution.

flamedisx/lxe_blocks/quanta_generation.py
robertsjames: This is just a simplification of the change made way back in #127, yes. Back then I had `extra_dimensions` and `bonus_dimensions` (the distinction there was only made to predict what I'd do in this PR, and I ended up simplifying things when it came to coding it up). The user must handle the dimsize/stepping calculation manually for any of these in `_calculate_dimsizes_special()` and tensor construction for them must be manually done in `_domain_dict_bonus()`. This currently happens here, and in `quanta_splitting.py` in the NEST sources. The flag just flags up whether or not the dimension represents a hidden variable that will be summed over in the final computation, and thus whether a corresponding variable stepping scaling should be applied or not. I've changed the comment a little in `block_surce.py` to make that clearer.

tests/test_inference.py
robertsjames: Done.

flamedisx/nest/lxe_blocks/quanta_splitting.py
robertsjames: Done.

flamedisx/nest/lxe_blocks/quanta_splitting.py
robertsjames: Done.

flamedisx/source.py
JelleAalbers: Ah ok, then could you use the `array_columns` feature to let flamedisx handle these 'tensors of columns' automatically? The WIMP energy spectrum block already does this, to have a different energy spectrum for each event (to account for modulation-induced spectral shifts), see e.g. https://github.com/FlamTeam/flamedisx/blob/main/flamedisx/lxe_blocks/energy_spectrum.py#L332. A complete neutron background model will have to use this feature as well. 

It will increase the data tensor's size a bit, but since the number of bounds ~ number of energies, and adding those is managable for the WIMP source, I don't think it's a show stopper.

flamedisx/TFP_files/truncated_skew_gaussian_cc.py
```diff
@@ -0,0 +1,178 @@
...
+  def _event_shape(self):
+    return tf.TensorShape([])
+
+  def _log_prob(self, x):
+    scale = tf.convert_to_tensor(self.scale)
+    skewness = tf.convert_to_tensor(self.skewness)
+    limit = tf.convert_to_tensor(self.limit)
+    bounded_log_prob = tf.where((x > limit),
+                                dtype_util.as_numpy_dtype(x.dtype)(-np.inf),
+                                tf.math.log(skew_gaussian.SkewGaussian(loc=self.loc,scale=scale,skewness=skewness,owens_t_terms=self.owens_t_terms).cdf(x+0.5) \
```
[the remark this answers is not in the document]
robertsjames: Done

flamedisx/TFP_files/truncated_skew_gaussian_cc.py
```diff
@@ -0,0 +1,178 @@
+# Copyright 2021 Robert James
+# ============================================================================
+"""The discretised Truncated Skew Gaussian distribution, with continuity correction, class."""
```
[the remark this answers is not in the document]
robertsjames: This worked! Thanks for the suggestion, I don't know why I didn't think of this!

flamedisx/source.py
robertsjames: Done!

robertsjames:
Okay, I've addressed everything now (bar the one thing we agreed could be left until later). I've asked @jordan-palmer to re-run the validation tests from our paper just to be sure nothing was broken in the process... Once we have that confirmed, I think we're ready to merge!

plt109:
I'll run some fits to the MC events generated from the default ER source as well just to check.

plt109 (changes requested):
So with these changes, the user would not be able to pass in `max_dim_size` and `max_dim_size_outer` as an input argument to `fd.LogLikelihood(..)`. This is because `max_dim_size_outer` does not exist anymore since `max_dim_size` can be tuned in each block by having it as a dictionary as seen [on this line](https://github.com/robertsjames/flamedisx/blob/eb047911d304f9955836faee7f1303541f821714/flamedisx/nest/lxe_blocks/quanta_splitting.py#L21) in `MakePhotonsElectronsNR` of the nest source.

To future-proof this per-block tuning of `max_dim_size`, I'm thinking of coding something similar explicitly in each block in the default sources. Could you include an assertion somewhere to check that this dictionary does not get overwritten by the different blocks?

robertsjames:
@jordan-palmer found a case which highlighted a bug - I've fixed it now. He's going to run those validation checks to make sure there's nothing else I've missed. I'm also going to address your comment now @plt109. @JelleAalbers - once Jordan confirms things are still okay from our NEST validations, and Pueh okays everything from the default source side, will you be ready to merge? We really want to get this paper on the arxiv!

JelleAalbers:
Thanks for the fixes Robert! I'm happy you were able to work around the `tfp` and  i_batch stuff.

Am I right that nobody has reviewed the new bounds stuff yet? I have not looked at it but I think someone should. Did you discuss this with someone at Stockholm, or should I have a look?

Could you describe the bug and fix? Since it's a change to the core BlockSource.

robertsjames:
The bounds stuff still only affects the NEST sources. The changes are detailed in the [slides](https://docs.google.com/presentation/d/1CZJYkNbn1ozxGKHkDhzFp6O9agFfZl6wRM8M2bPHmRY/edit#slide=id.g11d8e2e532e_0_47) and the paper, which we did go over in Stockholm. I'd hoped to spend some of the time there doing the more general Bayes bounds that leverage directly from compute(), so the default sources can benefit too, but this was swamped by the mu issue and other problems. It's still pretty high on the list of future work though. So given the scope of the bounds changes, not sure that it necessarily needs pouring over in too much detail right now. Up to you though.

As far as the bug goes, the problem was that the function signature to the TF graph was being computed before the data tensor size was known - the number of ion bounds inserted was determined from the data. I've changed it now so that it's set to be the maximum it can be (which is the minimum of the length of the energy spectrum and max_dim_sizes['energy']), and I pad the ion bounds with zeroes if necessary.

JelleAalbers (approved):
Ah OK. Am I right that the default sources don't use the Bayes bounds, and the MC_reservoir is never filled, so the stuff in get_priors is skipped?  Then I agree I don't need to review that part :-). Just two minor comments based on a quick glance:
  * Does self.MC_reservoir need to be both an attribute and passed as an argument to get_priors?
  * I would prefer `mc_reservoir` rather than `MC_reservoir`; uppercased things indicate classes to me.

Since you addressed all my other comments, I think we can merge once you and Pueh converge on the max_dimsize handling.

flamedisx/nest/lxe_blocks/quanta_splitting.py
```diff
@@ -81,17 +86,23 @@ def compute_single_energy_full(args):
...
+                else:
+                    normal_dist_ni = tfp.distributions.Normal(loc=nq_mean*alpha,
+                                                    scale=tf.sqrt(nq_mean*alpha) + 1e-10)
+                    p_ni = normal_dist_ni.cdf(_ions_produced + 0.5) - \
+                        normal_dist_ni.cdf(_ions_produced - 0.5)
+
+                    normal_dist_nq = tfp.distributions.Normal(loc=nq_mean*alpha*ex_ratio,
+                                                    scale=tf.sqrt(nq_mean*alpha*ex_ratio) + 1e-10)
+                    p_nq = normal_dist_nq.cdf(nq - _ions_produced + 0.5) - \
+                        - normal_dist_nq.cdf(nq - _ions_produced - 0.5)
```
jordan-palmer, line 106: double negative here

robertsjames:
@plt109, I've hard-coded the max_dim_sizes in for the default sources now, in the blocks

robertsjames:
> Ah OK. Am I right that the default sources don't use the Bayes bounds, and the MC_reservoir is never filled, so the stuff in get_priors is skipped? Then I agree I don't need to review that part :-). Just two minor comments based on a quick glance:
> 
> * Does self.MC_reservoir need to be both an attribute and passed as an argument to get_priors?
> * I would prefer `mc_reservoir` rather than `MC_reservoir`; uppercased things indicate classes to me.
> 
> Since you addressed all my other comments, I think we can merge once you and Pueh converge on the max_dimsize handling.

I've addressed this now

JelleAalbers:
Thanks Robert! @plt109 are you happy with the max_dim_sizes change? If so, I think it's time to merge this.

CristianAntochi:
So I tested multiple times trying to fit with the base sources and the XENON sources data and something is going wrong. If we keep `max_sigma=3` or just as default, the fit works. If one tries to increase the `max_sigma` in the likelihood to more than that the fit will not start, saying:
`ValueError: The likelihood is finite at your guess, but the gradient is not. Are you starting at a cusp?`
no matter what the guess is. I believe this is a bounds issue.

jordan-palmer:
> So I tested multiple times trying to fit with the base sources and the XENON sources data and something is going wrong. If we keep `max_sigma=3` or just as default, the fit works. If one tries to increase the `max_sigma` in the likelihood to more than that the fit will not start, saying: `ValueError: The likelihood is finite at your guess, but the gradient is not. Are you starting at a cusp?` no matter what the guess is. I believe this is a bounds issue.

jordan-palmer:
> So I tested multiple times trying to fit with the base sources and the XENON sources data and something is going wrong. If we keep `max_sigma=3` or just as default, the fit works. If one tries to increase the `max_sigma` in the likelihood to more than that the fit will not start, saying: `ValueError: The likelihood is finite at your guess, but the gradient is not. Are you starting at a cusp?` no matter what the guess is. I believe this is a bounds issue.

I tried to reproduce this but I cant, can you provide a minimum example please?

jordan-palmer:
> > So I tested multiple times trying to fit with the base sources and the XENON sources data and something is going wrong. If we keep `max_sigma=3` or just as default, the fit works. If one tries to increase the `max_sigma` in the likelihood to more than that the fit will not start, saying: `ValueError: The likelihood is finite at your guess, but the gradient is not. Are you starting at a cusp?` no matter what the guess is. I believe this is a bounds issue.
> 
> I tried to reproduce this but I cant, can you provide a minimum example please?

Okay, @plt109 informed me that you get this issue using the SR1ERSource which I can't run. I was getting correct results using the ERSource though

robertsjames:
Cristian sent me the code to produce the error. I'm going to try and dig into it later today. I agree we probably shouldn't merge until this is addressed

jordan-palmer:
Apologies for closing, I missed the button I wanted to press...

flamedisx/source.py
```diff
@@ -32,8 +32,12 @@ class Source:
     trace_difrate = True
 
     default_max_sigma = 3
+    default_max_sigma_outer = 3
     default_max_dim_size = 70
-    default_max_dim_size_outer = 120
+
+    # Capping the domain size for hidden variable dimensions. Any which aren't
+    # set will default to default_max_dim_size
+    max_dim_sizes: ty.Dict[str, int] = dict()
```
CristianAntochi, line 40: Can you add a comment with how to give the `max_dim_sizes` for all the blocks both in the NEST sources and in the standard sources? 
An example of a dict would suffice.

robertsjames:
@plt109 @CristianAntochi I think the problem with the fits was simply that I hadn't rebased this branch with main (this happens automatically when merging), so by checking the branch out and doing a fit lots of the Xenon1T files reverted to older versions. I've rebased now, and repeating your fit for the `SR1ERSource` now gives the same results as the main branch. Does one of you want to confirm this? Then we can merge

CristianAntochi (approved):
Now Fd seems to be fitting with the XENON sources as well. It was indeed a problem of rebasing. Sorry for not spotting this earlier.

plt109 (approved):
Looks good!

Issue #194, opened 2022-01-18
Restructuring of BBF data files has broken SR1 Xenon model


Ashley-Joy:
BBF has restructured its /bbf/data folder, mostly placing files required by the SR1 model in xenon/x1t_sr1.py into various subdirectories, breaking use of the models in that file. Configuration files can override these but the defaults should also be made to work.

I've implemented a quick fix in the 'fixBBF' branch by altering the paths.

CristianAntochi:
Hey Ash, I think you can commit it to my PR #193  if @plt109 and the others are ok with it, since it is just a path change, no? I will also change the electron detection efficiency in the same PR, being a super easy fix.

Ashley-Joy:
That's correct, it's just path changes. It looks a little messy since the paths are now much longer (they used to be in the same folder, so the path to that was stashed away in the "get_bbf_file" function) so there could be some cleanup with string addition, but they're still spread over 3 different folders now so there's not too much to do.

plt109:
Hello Ash, I would rather you make a separate PR from @CristianAntochi's PR 193. We should stick to having only 1 change per PR. I will review it when you're done.

plt109:
Actually, how many files were affected? A quick fix is to place the affected files in the flamedisx nT maps repo (https://github.com/XENONnT/Flamedisx/tree/master/nt_maps), 1T maps repo (https://github.com/XENONnT/Flamedisx/tree/master/1t_maps). Then we can do away with the `get_bbf_file` function all together.

The maps are usually not too big so this duplication does not waste too much space, and then we wouldn't be affected by any future restructuring of the bbf folders.

PR #209, opened 2022-03-23
path update for maps for x1t sources

CristianAntochi:
Urgent PR to update the paths for the x1t sources.


CristianAntochi:
Not super needed since it can be overwritten by the config file, but still, good to have working paths also in the defaults of the sources, without fishing for the correct path.

CristianAntochi:
This fixes issue #194

plt109:
Looks good! Even though the paths would be overwritten in the config file but it's good to have the sources working even if one were to not provide a config file.

plt109 (approved):

PR #213, opened 2022-03-28
Add field to config

plt109:
The default drift field is made a model attribute so that it can be set via a configuration file instead of changing the value of `DEFAULT_DRIFT_FIELD` in `x1t_sr1.py`. This is useful to adapt the current drift field value which was set to the field value of xenon1T SR1 to the nominal drift field value in xenonnT. Note that the drift field is still a single value throughout the entire TPC (drift field is not a function of position).

The changes in this PR affect both `fd.SR1ERSource` and `fd.SR1NRSource` via `p_electron` in each of these sources. To ensure that the changes do not induce any bugs in the fitting process, MC events were simulated and fitted for both `fd.SR1ERSource` and `fd.SR1NRSource`.

For `fd.SR1ERSource`, the optimizer converged as expected (the biases are still there, but this is a known feature) when floating the full set of ER model parameters.
Screenshot from 2022-03-28 17-06-05

For `fd.SR1NRSource`, the optimizer converged as expected (the biases are still there, but this is a known feature) when floating the NR model parameters except Lindhard K, which @Ashley-Joy does not intend to float from his other studies.
Screenshot from 2022-03-28 17-06-14

Testing notebook available upon request.

### DIFF
--- a/flamedisx/xenon/x1t_sr1.py
+++ b/flamedisx/xenon/x1t_sr1.py
@@ -422,7 +422,12 @@ class SR1NRSource(SR1Source, fd.NRSource):
         """
         # TODO: so to make field pos-dependent, override this entire f?
         # could be made easier...
-        #drift_field = tf.constant(self.default_drift_field, dtype=fd.float_type())
+
+        # in _compute, n_events = batch_size
+        # drift_field is originally a (n_events) tensor, nq a (n_events, n_nq) tensor
+        # Insert empty axis in drift_field for broadcasting for tf to broadcast over nq dimension
+        if tf.is_tensor(nq):
+            drift_field = drift_field[:, None]
 
         # prevent /0  # TODO can do better than this
         nq = nq + 1e-9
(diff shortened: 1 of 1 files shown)


CristianAntochi:
I would also give the possibility of reading the drift field from a map in the same PR. This still sets the drift field as a constant, no? 
One idea would be to implement it through a flag? or through a check if the drift field is given as a map, or a map path, and if not, use the fixed config.

plt109:
This PR is meant to be small and quick just so that I can come up with the config file for nT which can be used by the other xenon flamedisx folks.

Reading in the full drift field map as a function of (x,y,z) would be accompanied by the associated field distortion map, distortion correction map and pos-rec smearing functions which are in the [add_fdc branch](https://github.com/FlamTeam/flamedisx/tree/add_fdc) but not ready to be merged yet.

Ashley-Joy (approved):

Ashley-Joy:
I agree with Pueh on this. Loading from a map is a much larger project that is lower in priority. We are using a fixed value right now, and it's important we be able to adjust this through a config file (or at least not through recompiling code). This will all get overwritten later, but not soon enough that this change isn't worthwhile.

PR #217, opened 2022-04-13
Add aft to config

plt109:
The default s2 area fraction top is made a model attribute, `s2_area_fraction_top`, so that it can be set via a configuration file instead of changing the value of `DEFAULT_AREA_FRACTION_TOP` in x1t_sr1.py. It is also explicitly called `s2_area_fraction_top` to differentiate it from `s1_area_fraction_top` which is used in other cuts currently not explicitly used in flamedisx.

This is useful to change the s2 area fraction top value which is used to compute the cs2b values in `s2_acceptance` and apply the cuts on cs2b.

The changes in this PR affect both fd.SR1ERSource, fd.SR1NRSource, fd.SR1WallSource via `s2_acceptance` from the base class `SR1ERSource`. To ensure that the changes do not induce any bugs in the fitting process, MC events were simulated and fitted for both fd.SR1ERSource and fd.SR1NRSource.

For fd.SR1ERSource, the optimizer converged as expected (the biases are still there, but this is a known feature) when floating the full set of ER model parameters.
Screenshot from 2022-04-13 19-33-16

For fd.SR1NRSource, the optimizer converged as expected (the biases are still there, but this is a known feature) when floating the NR model parameters except Lindhard K, which @Ashley-Joy  does not intend to float from his other studies.
Screenshot from 2022-04-13 19-28-19

Testing notebook available upon request.


Ashley-Joy (approved):

plt109:
Decided to delete this PR because of the messy commit history and merge in the changes from PR 218 instead.

PR #221, opened 2022-05-08
Add field map

plt109:
## Why?
The spatial variation of the drift field gets increasingly important as liquid xenon TPCs become larger. The ability to model such inhomogeneities in the drift field also opens up the possibility of extending the fiducial volume instead of trimming those regions away.

## What? How?
This pull request allows the user to incorporate a spatially dependent drift field into `SR1ERSource` and `SR1NRSource` which is interpolated from an external drift field map. The field distortion and field distortion correction are also included.

## Testing? Screenshots (optional)
The fit to ER MC events converges without any issue even though the bias is still there:
Screenshot from 2022-05-09 00-30-58
The fit to NR MC events converges without any issue even though the bias is still there:
Screenshot from 2022-05-09 00-59-46
@Ashley-Joy, you might want to test this to make sure that the changes do not cause any breaking changes to your work. The drift field can be set to a single number(turn off spatial dependence of drift field) by setting `variable_drift_field = False` in the config file.

### DIFF
--- a/flamedisx/utils.py
+++ b/flamedisx/utils.py
@@ -280,7 +280,7 @@ def load_config(config_files=None):
             if not k.startswith('_')})
 
     return config
-    
+
 
 # Taken from straxen to filter arguments for interpolators
 @export
--- a/flamedisx/utils.py
+++ b/flamedisx/utils.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 import subprocess
 
+import inspect
 import numpy as np
 import pandas as pd
 from scipy import stats
--- a/flamedisx/utils.py
+++ b/flamedisx/utils.py
@@ -279,3 +279,17 @@ def load_config(config_files=None):
             if not k.startswith('_')})
 
     return config
+    
+
+# Taken from straxen to filter arguments for interpolators
+@export
+def filter_kwargs(func, kwargs):
+    """Filter out keyword arguments that
+        are not in the call signature of func
+        and return filtered kwargs dictionary
+    """
+    params = inspect.signature(func).parameters
+    if any([str(p).startswith('**') for p in params.values()]):
+        # if func accepts wildcard kwargs, return all
+        return kwargs
+    return {k: v for k, v in kwargs.items() if k in params}
--- a/flamedisx/xenon/itp_map.py
+++ b/flamedisx/xenon/itp_map.py
@@ -199,7 +199,7 @@ class InterpolatingMap:
         assert dimensions == 2, 'RectBivariateSpline interpolate maps of dimension 2'
         assert not array_valued, 'RectBivariateSpline does not support interpolating array values'
         map_data = map_data.reshape(*grid_shape)
-        kwargs = straxen.filter_kwargs(RectBivariateSpline, kwargs)
+        #kwargs = straxen.filter_kwargs(RectBivariateSpline, kwargs)
         rbs = RectBivariateSpline(grid[0], grid[1], map_data, **kwargs)
 
         def arg_formated_rbs(positions):
@@ -221,7 +221,7 @@ class InterpolatingMap:
             map_data = map_data.reshape(*grid_shape)
 
         config = dict(bounds_error=False, fill_value=None)
-        kwargs = straxen.filter_kwargs(RegularGridInterpolator, kwargs)
+        #kwargs = straxen.filter_kwargs(RegularGridInterpolator, kwargs)
         config.update(kwargs)
         
         return RegularGridInterpolator(tuple(grid), map_data, **config)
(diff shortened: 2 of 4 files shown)


flamedisx/utils.py
```diff
@@ -279,3 +280,17 @@ def load_config(config_files=None):
             if not k.startswith('_')})
 
     return config
+
+
+# Taken from straxen to filter arguments for interpolators
+@export
+def filter_kwargs(func, kwargs):
```
CristianAntochi, line 287: There is a similar function for the sources `_filter_source_kwargs`, but it is good to have a general function for that.

flamedisx/xenon/itp_map.py
```diff
@@ -109,58 +122,149 @@ def __init__(self, data):
...
+            elif method == 'WeightedNearestNeighbors':
+                itp_fun = self._weighted_nearest_neighbors(csys, map_data, array_valued, **kwargs)
+
             else:
-                if array_valued:
-                    map_data = map_data.reshape((-1, map_data.shape[-1]))
-                itp_fun = InterpolateAndExtrapolate(points=np.array(cs),
-                                                    values=np.array(map_data),
-                                                    array_valued=array_valued)
+                raise ValueError(f'Interpolation method {method} is not supported')
```
CristianAntochi, line 185: In case of error, could you add the supported methods in the error message?

flamedisx/xenon/x1t_sr1.py
```diff
@@ -200,6 +204,16 @@ class SR1Source:
     variable_elife = True
     path_electron_lifetimes = ('1t_maps/electron_lifetimes_sr1.json',)
 
+    # Comsol map
+    variable_drift_field = True
+    path_drift_field = 'nt_maps/fieldmap_2D_B2d75n_C2d75n_G0d3p_A4d9p_T0d9n_PMTs1d3n_FSR0d65p.json'
```
CristianAntochi, line 209: nT map path for an 1T source, with a default `variable_drift_field = True`. To keep the 1T sources viable without config files you should switch off the default to `False`.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -200,6 +204,16 @@ class SR1Source:
...
+
+    # Field distortion map
+    path_drift_field_distortion = 'nt_maps/init_to_final_position_mapping_B2d75n_C2d75n_G0d3p_A4d9p_T0d9n_PMTs1d3n_FSR0d65p.json'
+
+    # FDC map
+    path_drift_field_distortion_correction = 'nt_maps/XnT_3D_FDC_xyt_MLP_v0.2_B2d75n_C2d75n_G0d3p_A4d9p_T0d9n_PMTs1d3n_FSR0d65p.json'
```
CristianAntochi, line 215: This FDC is from nT and if you are to use a 1T source for an 1T analysis this would be wrong! Because you use it ...

flamedisx/xenon/x1t_sr1.py
```diff
@@ -251,11 +279,43 @@ def random_truth(self, n_events, fix_truth=None, **params):
...
+        # x, y, z are the true positions of the event
+        #
+        # x_observed, y_observed are the reconstructed positions with 
+        # field distortion and with posrec smearing but without 
+        # field distortion correction
+        #
+        # x_fdc, y_fdc, z_fdc are the fdc-corrected positions
+        
+        # going from true position to position distorted by field
+        d['r_observed'] = self.drift_field_distortion_map(
```
CristianAntochi, line 291: ... here, ...

flamedisx/xenon/x1t_sr1.py
```diff
@@ -251,11 +279,43 @@ def random_truth(self, n_events, fix_truth=None, **params):
...
+
+        # leave z intact, might want to correct this with drift velocity later
+        d['z_observed'] = d['z']
+
+        # Adding some smear according to posrec resolution
+        d['x_observed'] = np.random.normal(d['x_observed'].values, scale=0.4) # 4 mm resolution)
+        d['y_observed'] = np.random.normal(d['y_observed'].values, scale=0.4) # 4 mm resolution)
+        
+        # applying fdc
+        delta_r = self.fdc_map(
```
CristianAntochi, line 305: ... and here. My suggestion would be to keep things as they were for 1T and add a flag for nT where positions are corrected by the FDC maps.

CristianAntochi (changes requested):

flamedisx/xenon/x1t_sr1.py
plt109: I agree and I have just pushed the suggested change.

flamedisx/xenon/x1t_sr1.py
plt109: This is just a placeholder for the maps which the user should change.

flamedisx/xenon/itp_map.py
plt109: Interpolation is a delicate operation and I think it would be better for the user to safer for the user to be mindful of the exact interpolation algorithm. The error message in `ValueError` could be edited to include the supported interpolation algorithms but that would have to be changed each time an algorithm is included.

The interpolation functions are adapted from those in straxen and we expect to see more changes in the near future.

CristianAntochi (approved):

CristianAntochi:
If you say no one cares about 1T, merge!

Ashley-Joy (approved):

Issue #71, opened 2020-05-02
Correct rate parameter fit for wrong mu estimation


JelleAalbers:
In our paper, we argue that flamedisx's mu estimation doesn't have to be very good. As long as there are no strong external contraints on the rate of a source, fits will converge to correct values of the shape parameters. We can then correct the result for the rate parameters with a single high-statistics estimate of mu at the fitted shape parameters.

It would be good to add functionality to flamedisx to actually do this correction.

For limit setting, this means we have to return not just a limit, but also the (nuisance) parameters at which the limit is reached. If we do the correction internally we might not have to return them, but we should at least track them a little longer.

JelleAalbers:
If we want to allow grid interpolation, we could use this code: https://github.com/JelleAalbers/hypney/blob/main/hypney/utils/interpolation.py#L24. It's written for eagerpy so it will work with tensorflow (I ported some pytorch code referenced at the top, unit tests elsewhere in the repo). 

If we want unstructured interpolation, we could use scipy's LinearNDInterpolation to create a regular grid first, then use the code above.

Since the number of gridpoints is exponential in the number of parameters, we might want to give users to option to do grid interpolation just for a few parameters (with the others using star/cross interpolation as we do now).

Re-doing the mu estimate on each likelihood evaluation won't work, then the likelihood becomes non-differentiable. But we could try some iterative procedure:
  - Do the fit.
  - Estimate mu at the best-fit point, using simulation.
  - If interpolated mu is close enough, terminate.
  - Otherwise, add the best-fit point to the set of known mus (and perhaps estimate mu at a few other points around it too).
  - Build new interpolator, go back to step 1.

Issue #52, opened 2019-11-15
Cache system for mu computation and interpolation


CristianAntochi:
Since I'm working with Blueice now, I was thinking that we should start thinking of a smart way to implement a caching system for Flamedisx as well, in particular for the computation of the mus in the likelihood initialisation having in mind the big amount of toyMCs that we will need for the analyses.
We should think of a way to initialise and cache the mu computation and a function to just recall the precomputed mus. 
Related to this, to avoid fluctuations in the mu interpolation we should also think at a smart way to smoothen the interpolation so we don't create local minima that might give problems in the fitting of the likelihood function.

JelleAalbers:
Indeed: for the caching I think the blueice approach would work if we add Knut's fix in https://github.com/JelleAalbers/blueice/issues/29 to prevent race conditions. 

The interpolation issue is harder: we can try splining things or using a smoother interpolation (inverse weighted distance maybe?). The tricky thing is it has to be in tensorflow to work.

mayou36:
The interpolation sounds interesting, as we're ([zfit](https://github.com/zfit/zfit) also mostly based on TensorFlow and aiming at providing interpolation schemes to smooth binned functions.

Our main goal is though to move all the TF efforts together so that we don't need to rewrite all this functions in TF. Did you already start to do something in the interpolation direction?

Btw: if we wanna set this things up, it could be good to contribute it directly to [TensorFlow Probability](https://github.com/tensorflow/probability)

PR #222, opened 2022-05-12
Mu estimation options, saving and loading

JelleAalbers:
This expands flamedisx's options for estimating the total expected events ('mu'). It mostly prepares the ground for implementing new mu estimation methods in the near future, rather than adding new methods.

Specifying different mu estimation methods
----------------------------------------------------------
See issue #71. A new `mu_estimators` argument to `Likelihood.__init__` controls how mu estimators are built:
  * `fd.CrossInterpolatedMu` (the default) will use the current algorithm.
  * `fd.SimulateEachCallMu` does a completely new simulation each call. This does not actually work -- it makes the likelihood non-differentiable. But it could be useful for debugging, or for future developement (I think Pueh is looking into this)
  * `fd.ConstantMu` assumes mu is a constant. Perhaps useful to separate different contributions to the likelihood.
  * You can pass an already-built dictionary of mu interpolators, e.g. loaded from disk (see next section);
  * If you have multiple sources in the likelihood, and want to use different mu estimation methods for each, pass a dictionary mapping source names to the options above.
  * Finally, `fd.CombinedMu` uses different mu estimators for different parameter subsets. For example, this will setup two independent 2d grid interpolators, and use cross interpolation for the remaining parameters:
```
est = fd.CombinedMu.from_estimators(
    {('elife', 'g1'): fd.GridInterpolatedMu,
     ('g2', 'single_electron_size'): fd.GridInterpolatedMu},
    default=fd.CrossInterpolatedMu)
fd.LogLikelihood(..., mu_estimators=est)
```

The actual code for mu estimation now lives in a new file (`mu_estimation.py`) which contains classes for the different methods. Currently it's a short list, but we can start adding to this. `Source`s no longer have a `mu_function` method; they are only responsible for `estimate_mu` at one parameter combination.

This PR also generalizes the third element in parameter specifications. Currently this is a number of anchors:

```python
ll = fd.LogLikelihood(
    sources=dict(er=fd.ERSource),
    elife=(300e3, 500e3, 2),)
```

This remains supported for backwards compatibility. However, an anchor count is not meaningful for all mu estimation methods. Instead, you can omit the third element (fd.CrossInterpolation will use 2 anchors by default):

```python
ll = fd.LogLikelihood(
    sources=dict(er=fd.ERSource),
    elife=(300e3, 500e3),)
```

or pass a dictionary, in which you can put other options as well:

```python
ll = fd.LogLikelihood(
    sources=dict(er=fd.ERSource),
    elife=(300e3, 500e3, dict(n_anchors=2),)
```

Saving and loading mu estimators
-----------------------------------------------
See issue #52. Now that we're going to spend more computation time on the mu estimation, we don't want to redo that work every time we start a notebook.

You can set the new `mu_estimators` argument to a previously built mu_estimators dictionary. For example,

```python
ll = fd.LogLikelihood(....)

# Saving to disk
fn = 'bla.pkl'
with pickle.open(fn, mode='wb') as f:
    pickle.dump(f, ll.mu_estimators)
    
# Loading from disk
with pickle.open(fn, mode='rb') as f:
    mu_estimators = pickle.load(f)
ll = fd.LogLikelihood(...., mu_estimators=mu_estimators)
```

Maybe we want to build something more opinionated (a la blueice) on top of this later, or maybe simpler is better for now.

I would still like to add a few unit tests.

### DIFF
--- a/flamedisx/__init__.py
+++ b/flamedisx/__init__.py
@@ -6,6 +6,7 @@ from .block_source import *
 from .likelihood import *
 from .inference import *
 from .bounds import *
+from .mu_estimation import *
 
 # Original flamedisx models
 # Accessible under fd root package (for now), for backwards compatibility
--- a/flamedisx/block_source.py
+++ b/flamedisx/block_source.py
@@ -484,10 +484,10 @@ class BlockModelSource(fd.Source):
     def _simulate_response(self):
         # All blocks after the first help to simulate the response
         d = self.data
-        d['p_accepted'] = 1.
+        d['p_accepted'] = 1.   # Cut on p_accepted is made in Source.simulate
         for b in self.model_blocks[1:]:
             b.simulate(d)
-        return d.iloc[np.random.rand(len(d)) < d['p_accepted'].values].copy()
+        return d
 
     def get_priors(self, data):
         """Obtain priors on certain hidden variable dimensions, to obtain more
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -213,7 +213,7 @@ class LogLikelihood:
         self.mu_estimators = {}
         for sname, s in self.sources.items():
             mu_est = mu_estimators[sname]
-            if issubclass(mu_est, fd.MuEstimator):
+            if fd.is_mu_estimator_class(mu_est):
                 # Build a new mu estimator using the source and configuration
                 mu_est = mu_est(
                     source=s,
@@ -221,10 +221,12 @@ class LogLikelihood:
                     progress=progress,
                     # Source will filter out the params it needs
                     **common_param_specs)
+            elif isinstance(mu_est, fd.MuEstimator):
+                # This is an already-built estimator -- perhaps loaded
+                # from a pickle
+                pass
             else:
-                # Assume this is an already-built estimator -- perhaps loaded
-                # from a pickle -- or some other function we can call
-                assert callable(mu_est)
+                raise ValueError(f"Invalid mu estimator {mu_est}")
             self.mu_estimators[sname] = mu_est
 
         # Not used, but useful for mu smoothness diagnosis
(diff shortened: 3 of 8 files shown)


robertsjames:
Thanks so much for this @JelleAalbers - this will be very useful for testing the different options. I have a slightly manic couple of weeks ahead with other projects but when I get back to thinking about this I'll dump some of my code (currently in standalone Jupyter notebooks) into here

CristianAntochi:
This looks great! I also have some code to put here but it is based on Rob's work on the mu Taylor expansion. The idea is to have multiple Taylor expansions around different points in the parameter space and weight the functions according to the distance where the evaluation happens. When Rob puts in the Taylor expansion I could gladly put in my method as well.

JelleAalbers:
Allright, finally added units tests; now this is ready for review. If nobody gets to this in a while, I may add GridInterpolation -- but I would prefer to make that a separate PR, this is already a big one.

I added a fix to how flamedisx handles empty data (datasets with no events), which was useful for writing the mu estimator tests. Previously feeding in empty data would crash the source (during populate_tensor_cache) and cause trouble in the likelihood (the mu and constraint term get added on batch #0, but skipped if there are no batches).

jordan-palmer:
Thank you for doing this, I read through it a while ago but have had a more thorough look this morning.

I think having an argument in the Likelihood function is the best way of doing this and I like that the functions get their own file. 

I personally would like the grid interpolation method included in this PR or to quickly follow this PR (or another method than the cross interpolation). This would give us a good solution for uncorrelated and correlated parameters. Currently, we don't have a good option for correlated parameters. This would allow us to perform fits on the same data using both methods and maybe would inform us on the extent that this mu issue is impacting us. 

In terms of how this has all been implemented, I have no comments. It all looks neat to me.

robertsjames (approved):
This all looks great @JelleAalbers - sorry it took so long to take a look. Thanks for adding unit tests also, this saves me some work in checking things. I agree that merging this first before adding in any new methods is a good call. So feel free to merge this whenever you're ready!

JelleAalbers:
Thanks for your comments everyone! Let's merge this and start collecting new interpolation methods soon.

PR #223, opened 2022-05-13
Exposure in wimp source

plt109:
## What? Why?
This pull request introduces another class variable `exposure_tonneyear` to the `WIMPEnergySpectrum` class so that it is easier to change the exposure of the WIMP source.

## How?
The spectra computed by `wimprates.rate_wimp_std` is in the units of events / (keV tonne year) and it is scaled by the exposure in units of tonne year [in this line](https://github.com/FlamTeam/flamedisx/blob/d371b34db4de2256f3697a5e08d02e260ef40787/flamedisx/lxe_blocks/energy_spectrum.py#L358).

The way to scaling the WIMP rate is inspired by @pelssers's [sensitivity studies notebook](https://github.com/FlamTeam/flamedisx-notebooks/blob/b8acfdf8fcc2c191fe1882706466c442e46a92ed/flamedisx_sensitivity.ipynb). He tried to scale the spectrum from the class but there is a comment that suggests that that would not work because 'populate tensor cache is already called so has no effect' so in this PR, the scaling factor is introduced to `WIMPEnergySpectrum` instead.

## Testing? Screenshots (optional)
The changes are tested using the a class which inherits from `fd.SR1WIMPSource` and with the class variables overwritten as such:
```
class wimp_variable_f2(fd.SR1WIMPSource):
    variable_drift_field = True
    
    mw = this_mw
    sigma_nucleon = this_sigma_nucleon
    energy_edges = np.geomspace(0.7, 50., 100) # sticking to default values
    exposure_tonneyear = this_exposure
    
    #'''
    pretend_wimps_dont_modulate = False
    n_time_bins = 24
    #'''
    '''
    pretend_wimps_dont_modulate = True
    n_time_bins = 1
    #'''
```
Setting the exposure to 4 tonne * 10./365. years, the ~38 expected number of 50 GeV WIMP events(cross section 1e-45 cm2/nucleon) is obtained, which is equivalent to ~348 events, which is the number of expected WIMP events given 1 tonne year of exposure (default, no scaling).
Screenshot from 2022-05-13 19-08-44

### DIFF
--- a/flamedisx/lxe_blocks/energy_spectrum.py
+++ b/flamedisx/lxe_blocks/energy_spectrum.py
@@ -330,6 +330,7 @@ class WIMPEnergySpectrum(VariableEnergySpectrum):
     def setup(self):
         wimp_kwargs = dict(mw=self.mw,
                            sigma_nucleon=self.sigma_nucleon,
+                           exposure_tonneyear=self.exposure_tonneyear,
                            energy_edges=self.energy_edges)
 
         # BlockModelSource is kind enough to let us change these attributes
@@ -345,7 +346,8 @@ class WIMPEnergySpectrum(VariableEnergySpectrum):
 
         # Transform wimp_kwargs to arguments that can be passed to wimprates
         # which means transforming es from edges to centers
-        del wimp_kwargs['energy_edges']
+        del wimp_kwargs['energy_edges'], wimp_kwargs['exposure_tonneyear']
+
         spectra = np.array([wr.rate_wimp_std(t=t,
                                              es=e_centers,
                                              **wimp_kwargs)
--- a/flamedisx/lxe_blocks/energy_spectrum.py
+++ b/flamedisx/lxe_blocks/energy_spectrum.py
@@ -300,6 +300,7 @@ class WIMPEnergySpectrum(VariableEnergySpectrum):
     model_attributes = ('pretend_wimps_dont_modulate',
                         'mw',
                         'sigma_nucleon',
+                        'exposure_tonneyear',
                         'n_time_bins',
                         'energy_edges') + VariableEnergySpectrum.model_attributes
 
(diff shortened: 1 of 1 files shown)


JelleAalbers (approved):
Thanks Pueh, this looks good!

This simplifies the interpretation of the WIMP source's rate multiplier to a scale factor on the cross-section.

It may seem odd to have a exposure_tonneyear parameter for just the WIMPsource, while many backgrounds also scale with exposure. However:
- I don't think having something like a global rate multiplier would make sense, since not all backgrounds scale with exposure -- e.g. external backgrounds scale linearly with time but not with FV, while for some impurities it's the opposite. 
- In e.g. the ERSource you can set the entire energy spectrum by hand, and smuggle in the exposure scaling that way. However, in the WIMP source the spectrum is computed for you, so you had no good way of scaling the expectation until this PR. (E.g. adjusting the reference cross-section would be mightily confusing.)

plt109:
Thanks for looking through this PR Jelle!

Yups, I smuggled in the exposure scaling into the ER source through `rates_vs_energy` using the following:
```
fm_t = 4.2                                                                        # fiducial mass in tonnes
liveyears_yr = 42./365.                                                    # livetime in years
this_exposure_tyr = fm_t*liveyears_yr                             # exposure in tonne years
er_per_tyr = 42.                                                                # expected ER events per tonne year

num_expected_er = er_per_tyr*this_exposure_tyr           # expected ER events for given exposure

# Defining energy spectrum of ER band
max_er_energy = 20 # keV
num_energies = int(1000)

class er_fixed_f(fd.SR1ERSource):
    variable_drift_field = False
    default_drift_field = 42.
    energies = tf.cast(tf.linspace(0., max_er_energy, num_energies), dtype=tf.float32)
    rates_vs_energy = tf.ones(num_energies, dtype=tf.float32)*num_expected_er/num_energies
```

Issue #224, opened 2022-05-15
annotate_special/source.set_data: data tensor indexing edge case


jordan-palmer:
**Context**
I came across an issue using the `annotate_special` function.

When you call `source.set_data` and your cuts remove some number of events, we wish the data tensor with real events such that the total is a multiple of the batch size. In FD, we seem to use the 0th index, which is fine. This is done here
```                
self.data = pd.concat([self.data, df_pad], ignore_index=True)
```
where `df_pad` are the 0th event repeated some number of times. This line is only called if `n_padding` is > 0. In the case where `n_padding = 0` AND the case where the number of events cut are multiples of the batch size (10,20,30,.. in the default case) then this "ignore index" is not called and the final data tensor is awkwardly indexed. In the default annotate methods, this is does not cause an error, but this scenario does of course happen. Would be interesting to know if this could cause a feature somewhere? 

**Error message**

The result of this feature can break the `annotate_special` method with the following error:

```
/content/FlameNEST/flamedisx/nest/lxe_blocks/quanta_splitting.py in <listcomp>(.0)
    353         max_num_energies = max(min(len(self.source.energies), self.source.max_dim_sizes['energy']), 2)
    354         # Pad with 0s at the end to make each one the same size
--> 355         [bounds.extend([0]*(max_num_energies - len(bounds))) for bounds in d['ions_produced_min'].values]
    356 
    357         return True
AttributeError: 'float' object has no attribute 'extend'
```
After some researching, this error is due to this indexing issue. This occurs even with a fixed MC reservoir.

**Proposed solutions** 

- Update annotate_special to not break with awkward indexing
- if the `n_padding` = 0, just call `self.data = self.data.reset_index(drop=True)`. I don't believe this causes any issues in the logic from a couple of tests I have done. But this is by far the quickest solution and I always think it is good practise to reset the indices of dfs. 

**Recreate this error**

```
Likelihood = fd.LogLikelihood(sources = dict(er=fd.lux.LUXERSource),
                              free_rates = ('er'),
                              arguments = dict(er={'energy_min':0.01,'energy_max':20.}),
                             )
i=0
while i < 20:
    print('set data number ',i)
    df_data = fd.lux.LUXERSource(energy_min = 0.01, energy_max= 20.).simulate(200)
    Likelihood.set_data(df_data)
    i+=1
```

[...]

PR #225, opened 2022-05-16
Fix for git issue #224 

jordan-palmer:
Solution:
Simply reset index and drop original index for `self.data` in `source.set_data` such that the resulting data tensor is correctly indexed. This will allow the MC reservoir code to work but we could still think about updating it to be complete.


robertsjames:
Totally happy with this - thank so much for sorting it out @jordan-palmer!!

robertsjames (approved):

JelleAalbers (approved):

Issue #230, opened 2022-05-23
Bayes bounds errors when 0 in the energy spectrum


robertsjames:
The Bayes bounds break when the energy spectrum extends down to 0 energy. For now, avoid putting a 0 in your energy spectrum. I'll fix this as soon as I find some time

robertsjames:
Gave this a temporary fix in PR #231

PR #231, opened 2022-05-23
Fix lower default energies to avoid Bayes bounds problems.

robertsjames:
Temporary fix to issue #230


JelleAalbers (approved):
Sorry for the late response -- obviously this is fine. It only changes the NEST sources, and having 0-energy events doesn't really make sense anyway.

PR #241, opened 2022-07-27
Add simple fixed template source

JelleAalbers:
This allows users to specify sources using histograms that contain expected events/bin or differential rates. This is useful for sources like accidental coincides that have empirical models, and for cross-checks on the main flamedisx computation. Here's an example:
```python
source = fd.TemplateSource(some_multihist, events_per_bin=True, batch_size=int(1e3))
```
where `some_multihist` is a [multihist](https://github.com/JelleAalbers/multihist) Histdd. You should also be able to pass a [boost-histogram](https://github.com/scikit-hep/boost-histogram) or [hist](https://github.com/scikit-hep/hist) object, though I haven't tested this. Or you can just pass `template`, `bin_edges`, `axis_names` explicitly, setting the former two to numpy arrays and the final one to a sequence of strings. 

For a more complete example, see the [unit test](https://github.com/FlamTeam/flamedisx/blob/template_source/tests/test_template.py#L8).

When using a TemplateSource in a likelihood, you'll want to pass the template in via the `arguments` argument to LogLikelihood, or sneak it in via `functools.partial`. I haven't tested this though. We should probably make an example/tutorial that uses templates.

`TemplateSource` can simulate events, but it only simulates for dimensions that the histogram has. For example, a template source with a (cs1, cs2) histogram won't magically simulate raw (s1, s2) or uniform positions and times. If desired, this should be easy to add by making a TemplateSource subclass, and code like [here](https://github.com/FlamTeam/flamedisx/blob/main/flamedisx/lxe_blocks/energy_spectrum.py#L58).

A basic `TemplateSource` has no parameters, though if used in a likelihood you can fit its rate multiplier. In the future we can enable shape parameters / linear template morphing with [tfp.math.batch_interp_rectilinear_nd_grid](https://www.tensorflow.org/probability/api_docs/python/tfp/math/batch_interp_rectilinear_nd_grid) or [tfp.math.batch_interp_regular_nd_grid](https://www.tensorflow.org/probability/api_docs/python/tfp/math/batch_interp_regular_nd_grid) from tensorflow-probability.

### DIFF
--- a/flamedisx/__init__.py
+++ b/flamedisx/__init__.py
@@ -3,6 +3,7 @@ __version__ = '2.0.0'
 from .utils import *
 from .source import *
 from .block_source import *
+from .templates import *
 from .likelihood import *
 from .inference import *
 from .bounds import *
--- a/flamedisx/templates.py
+++ b/flamedisx/templates.py
@@ -1,4 +1,6 @@
-import tensorflow as tf
+import random
+import string
+
 from multihist import Histdd
 import pandas as pd
 
@@ -8,7 +10,7 @@ export, __all__ = fd.exporter()
 
 
 @export
-class TemplateSource(fd.Source):
+class TemplateSource(fd.ColumnSource):
     """Source that looks up precomputed differential rates in a template
     (probably a histogram from a simulation).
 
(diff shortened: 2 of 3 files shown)


robertsjames:
Hey Jelle, this looks great! Hope you're okay for me to go ahead and merge this - it shouldn't be a problem, as this is completely standalone code that shouldn't affect anything else

robertsjames (approved):
Hey Jelle, this looks great! Hope you're okay for me to go ahead and merge this - it shouldn't be a problem, as this is completely standalone code that shouldn't affect anything else

plt109:
Hello Jelle,

This is a really nice source and I briefly tested it and it seems to be fitting (codes running).
Screenshot from 2022-08-03 17-57-23

Do you think it's possible for Flamedisx to be competitive against blueice in terms of running time?

JelleAalbers:
Thanks Robert and Pueh!

> Do you think it's possible for Flamedisx to be competitive against blueice in terms of running time?

Yes. We'd  have to add shape parameter morphing first, but that should be possible with the tfp interpolator. Blueice doesn't use the derivative in the optimization, so I think a flamedisx likelihood with only template sources would be faster as soon as you have enough shape parameters.

PR #242, opened 2022-07-29
Rj grid interpolation

robertsjames:
This PR adds a grid interpolation option. The default number of anchors for a grid interpolated parameter is 3; cross interpolation automatically probes the centre in a roundabout way, but a 2-anchor grid interpolation would not, hence this choice.

As a benchmark, I did a fit with the Xe1T SR1 ER source (@plt109), for g1, gamma_er and omega_er. I used 2 anchors (plus the base mu at the centre) for grid interpolation, and 3 anchors for cross interpolation. Each anchor point had 10^6 simulated events to estimate mu.

The percentage differences between the interpolated and directly estimated mu in each case, at 3 benchmark points, are:

**Cross:**

Bestfit: 7.1%
Lower range for all: 9.4%
Central defaults for all: 4.4%

**Grid:**

Bestfit: 0.13%
Lower range for all: 0.22%
Central defaults for all: 0.045%

I added a couple of unit tests. I still want to test combining cross and grid interpolators.

### DIFF
--- a/flamedisx/likelihood.py
+++ b/flamedisx/likelihood.py
@@ -54,7 +54,7 @@ class LogLikelihood:
             batch_size=10,
             max_sigma=None,
             max_sigma_outer=None,
-            n_trials=int(1e5),
+            n_trials=None,
             log_constraint=None,
             bounds_specified=True,
             progress=True,
--- a/flamedisx/mu_estimation.py
+++ b/flamedisx/mu_estimation.py
@@ -239,23 +239,43 @@ class GridInterpolatedMu(MuEstimator):
     """Linearly interpolate the estimated mu on an n-dimensional grid"""
 
     def build(self, source: fd.Source):
-        _iter = self.bounds.items()
-        if self.progress:
-            _iter = tqdm(_iter, desc="Estimating mus")
+        _iter = dict(sorted(self.bounds.items())).items()
+
+        param_lowers = []
+        param_uppers = []
+        for pname, (start, stop) in _iter:
+            param_lowers.append(start)
+            param_uppers.append(stop)
+
+        self.param_lowers = param_lowers
+        self.param_uppers = param_uppers
 
         grid_dict = {pname: np.linspace(start, stop, int(self.param_options.get(pname, {}).get('n_anchors', 2))) for pname, (start, stop) in _iter}
         param_grid = list(ParameterGrid(grid_dict))
 
+        if self.progress:
+            param_grid = tqdm(param_grid, desc="Estimating mus")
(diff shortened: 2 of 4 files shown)


JelleAalbers:
This is great Robert, thanks! I only have a few minor tweaks, made a small PR-to-PR [here](https://github.com/robertsjames/flamedisx/pull/60) save us both time. Happy to discuss if you disagree or would like to clarify something.

JelleAalbers:
Just copying Robert's [comment](https://github.com/robertsjames/flamedisx/pull/60#issuecomment-1203845737) here:


> Hey Jelle - thanks for this! Just a couple of comments/questions:
> 
>  1. If the user now passes a different set of arguments to the grid interpolator than the arguments it was built with, what will happen?
>  2. WRT changing to 2 anchors instead of 3: I was seeing poor performance in this case, sometimes worse than the cross interpolation, due to the centre not being probed in any way. Are you sure this is a good idea? 
> We could increase the default stats. in this case to 10^6 perhaps, to avoid the cusps you mention. I'm just concerned that an unsuspecting user may get tripped up by assuming the grid interpolator works better, when with 2 anchors only this may not be the case...

Thanks for having a look! These are good points.

1. `MuEstimator.__call__` might be useful for power users, but I'd expect expect `LogLikelihood.mu` to be used in normal circumstances (since it filters params & fills defaults for you). That said, I think the behavior is the same as for the other mu estimators:
  * Spurious parameters are ignored;
  * Missing parameters get you an error (a KeyError in this case);
  * The parameter order does not matter (as your test verifies).
See also [here](https://github.com/FlamTeam/flamedisx/blob/77f7704437fbc4c2c202b5823386a41464d660d6/flamedisx/mu_estimation.py#L60).

2. Interesting!
  * Yes, GridInterpolator probes points further from the center than the CrossInterpolator, so using 2 anchors in the same range could perform worse in some situations. It depends on whether nonlinearities are stronger than correlations: in the first case a CrossInterpolator will do better.
  * Doing more sims gets you closer to the truth, but there will still be a discontinuity in the derivative. This shouldn't scare robust minimizers, but at least those heavily reliant on derivative/Hessian info may get confused. 
  * One other argument for 3 anchors however might be that you only use a GridInterpolator if you think cross isn't accurate enough and you want something beefier / are prepared to pay a price for that.

Anyone else have thoughts on 2 vs 3 anchors as default? If not, I would defer to Robert; I'm not actually working with this in practice at the moment.

robertsjames:
@JelleAalbers or @plt109, requesting another review

PR #247, opened 2022-08-08
Add FrozenReservoirSource

robertsjames:
This PR is to help things along with the work @plt109 and I are doing, generating TS distributions with toy MCs.

It allows you to use column sources to read off the differential rates from a reservoir containing a large number of events from each source in your analysis, with the differential rate of each event computed and stored under the model of each source. This will only work when you are just floating the rate multipliers.

You'd use it like this:

`reservoir = fd.source.FastSourceReservoir(sources=dict(er=fd.ERSource, wimp=fd.WIMPSource), arguments = dict(er={'batch_size': 100}, wimp={'batch_size': 100}), ntoys=1000)`
                                                                           
`ll_fast = fd.LogLikelihood(sources=dict(er=fd.FastSource, wimp=fd.FastSource), arguments = dict(er={'source_type': fd.ERSource, 'source_name': 'er', 'reservoir': reservoir}, wimp={'source_type': fd.WIMPSource, 'source_name': 'wimp', 'reservoir': reservoir}), progress=False, batch_size = 100, free_rates=('er', 'wimp'))`

Simulating/setting data from the likelihood and doing fits will then lead to rate multiplier fits happening very quickly. `ntoys` refers to how many toy experiments you will do to get your TS distributions; this then ensures the reservoir is sufficiently populated.


robertsjames:
Note: I also added some unit tests

JelleAalbers (commented):
Thanks Robert! This is a nice addition to enable faster rate multiplier fits. Couple suggestions:

  * Perhaps name the source `FrozenReservoirSource`? `FastSource` is really generic, other sources can also be fast.
  * For the `FastSourceReservoir` function, how about:
    * Call it e.g. like `make_event_reservoir`
    * Have it take source instances, not source classes and arguments. Currently the function instantiates sources twice, once with and once without arguments (?). 
    * You could even have it take **sources as argument, so users can just call e.g. `make_event_reservoir(er=fd.ERSource(), ...)`.
  * Maybe move this to short a new file (`frozen_reservoir.py`?). `source.py` is pretty full already.

robertsjames:
I've addressed these comments now @JelleAalbers. Can you please take another look?

PR #258, opened 2022-09-08
Allow for non-integer dimensions.

robertsjames:
This PR enables support for dimensions which are non-integer: this will be especially useful for @plt109's (and later our) reconstruction bias.

Using it is simple: I made a toy example where the final dimension becomes `s1_smear`, giving a Gaussian smearing `s1_smear ~ Normal(s1, 0.01)`. Once this new block has been implemented, and the correct new final dimensions have been flagged up (as before), the only additional change needed is to add

`non_integer_dimensions = ('s1')`

in the source class.

I've attached a validation plot for this model (using a LUX ER source at 10 keV with this new S1 smearing block bolted on, and a `max_dim_size` of 200 for `s1`). I attach that below

quantitative_comparison_er

### DIFF
--- a/flamedisx/block_source.py
+++ b/flamedisx/block_source.py
@@ -32,6 +32,9 @@ class Block:
     #: _calculate_dimsizes_special().
     bonus_dimensions: ty.Tuple[ty.Tuple[str, bool]] = tuple()
 
+    #: inner_dimensions which take non-integer values
+    non_integer_dimensions: ty.Tuple[str] = tuple()
+
     #: Columns that we don't want to include in the tensor of data columns
     exclude_data_tensor: ty.Tuple[str] = tuple()
 
@@ -251,6 +254,7 @@ class BlockModelSource(fd.Source):
         # Collect attributes from the different blocks in this dictionary:
         collected = {k: [] for k in (
             'dimensions',
+            'non_integer_dimensions',
             'bonus_dimensions',
             'exclude_data_tensor',
             'model_functions',
@@ -326,6 +330,8 @@ class BlockModelSource(fd.Source):
                 continue
             setattr(self, k, getattr(self, k) + tuple(set(v)))
 
+        self.non_integer_dimensions = tuple([
+            d for d in collected['non_integer_dimensions']])
         self.inner_dimensions = tuple(
             [d for d in collected['dimensions']
                 if ((d not in self.final_dimensions)
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -58,6 +58,9 @@ class Source:
     #: for which the domain is always a single interval of integers
     no_step_dimensions: ty.Tuple[str] = tuple()
 
+    #: inner_dimensions which take non-integer values
+    non_integer_dimensions: ty.Tuple[str] = tuple()
+
     #: Names of dimensions of hidden variables for which
     #: dimsize calculations are NOT done here (but in user-defined code)
     #: but for which we DO track _min and _dimsizes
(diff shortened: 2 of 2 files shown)


JelleAalbers (commented):
Thanks for this nice short addition Robert! Yet another type of `_dimensions` 🙃

I would suggest:
  * Allow blocks to specify which dimensions should be non-integer by adding your new attribute to the list here https://github.com/FlamTeam/flamedisx/blob/main/flamedisx/block_source.py#L254. That way you can more easily switch between blocks that do or don't use this.
  * After that, could you add some kind of test to show this is working? If you don't have time for that, maybe you could share your updated example somewhere (flamedisx_notebooks?), and/or maybe someone else (Pueh) can confirm this works for them?

robertsjames:
Thanks Jelle! Sorry it took a while to get back to this, but now Pueh wants to use it, so seemed like a good time! I address your suggestion. I believe @plt109 is now gonna test it out...

JelleAalbers:
Great, thanks Robert! If Pueh gets this working no objection to merge from me.

plt109:
Yups, I'm testing this functionality out now. The objective of this functionality is to allow us to add in the reconstruction bias block since to get from true s1 (or raw s1 depending how on you want to call it) to observed s1 (reconstructed s1), we will have to step on (sum over -> integrate over) non-integer steps (true s1). And also the same applies for s2's.

A test that this non-integer dimension works is that flamedisx is still able to construct the likelihood and fit for some parameters (rates, shape parameters etc) even if the reconstruction bias block is a dirac delta function (ok, maybe not so extreme that the gaussian has zero std, not sure yet).

I'll provide a simple notebook that works with `fd.ERSource()` and passes the above-mentioned test, if I get there.

plt109:
I was testing with `fd.ERSource` and I realised that I was naive to think that it'll be possible to have zero smearing for the last block and have any sensible `_compute` and `_simulate` comparison via some analysis similar to that in the [template accuracy notebook](https://github.com/FlamTeam/flamedisx-notebooks/blob/b8acfdf8fcc2c191fe1882706466c442e46a92ed/template_accuracy.ipynb).

The solution is to have a minimal smearing, which I will base on the amount of smearing we have in nT. However, that requires me to increase `max_dim_size` for the particular block. Looking at just s2, the `max_sigma` that I was using was 6, the WIMP search goes up to s2~2e4 PE, and this means that `max_dim_size` for the unsmeared s2 will have to be around 1700, which will make any computation infeasible in terms of memory usage.

image

@robertsjames Is it possible to add in a feature to tune the `max_sigma` for any block that uses this non-integer stepping functionality? This will not be a problem if you have automatically tuned bounds like the Bayes Bounds, but that itself poses challenges like the choice of priors and the MC reservoir that the Bayes Bounds depends on. As an interim solution, it'll be best to have blocks that uses this non-integer stepping feature to also have it's own `max_sigma`, pretty much like how this feature already requires the user to set the `non_integer_dimension` class attribute to step on non-integers.

plt109 (changes requested):
Being able to 'integrate' over non-integer values for the internal dimension allows us to add in the reconstruction bias block for both the default `ERSource` and `nest` sources. I tested this non-integer stepping functionality by checking the agreement between `_compute` and `_simulate` at just 2.8 keV for starters. Sticking to the default value of `max_sigma=3`, I got the following plot:
image
which indicates that `_compute` and `_simulate` do not agree if we just stick to `max_sigma=3`, but this is a known issue and I increased `max_sigma=6` and obtained the following plot:
image
which got rid of that weird upward-flicking tail, but there is clearly an offset, but this is also a known issue because `max_sigma=6` is too big for the block that steps on non-integer values. Sticking to `max_sigma=6` for all the blocks besides the block that steps on non-integer values ( `max_sigma=3`, `max_dim_size=120` for the non-integer block), I obtained the following plot.
image
This shows that so far, just testing on a mono-energetic blob at 2.8 keV using `ERSource`, this non-integer stepping functionality is not giving features that are not expected, yet.

However, since the tuning of `max_sigma` is so vital when one is stepping on non-integers, it would be also important to allow the user to tune the `max_sigma` of the non-integer block, similar to how the user would have to append the non-integer internal dimensions to the `non_integer_dimensions` tuple.

robertsjames:
@plt109 @JelleAalbers shall we merge this? I addressed the issue Pueh highlighted in https://github.com/FlamTeam/flamedisx/pull/272

JelleAalbers:
If Pueh agrees, sure -- in general if you two agree no need to wait on me :-)

plt109:
Ideally, we should rebase to include the features in [PR272](https://github.com/FlamTeam/flamedisx/pull/272) into this feature, but this PR has been here for a long time so I guess we can merge it in now.

plt109 (approved):
Look good. Let's go.

PR #261, opened 2022-09-18
Add S2 threshold properly, as NEST does.

robertsjames:
Very minor change to add in explicit S2 threshold, separate to S2 analysis cuts. Motivated by work being done for LZ fits - just thought I would add also to the default NEST (LUX) source...

### DIFF
--- a/flamedisx/nest/config/default.ini
+++ b/flamedisx/nest/config/default.ini
@@ -29,12 +29,14 @@ gas_gap_config = 4.25 ; mm
 g1_gas_config = 0.1
 s2Fano_config = 3.6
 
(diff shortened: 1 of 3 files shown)


JelleAalbers:
Hi Robert -- sorry, I'm not familiar with the NEST sources. Does this second S2 minimum/threshold setting do exactly the same as the existing one? Then I don't understand the motivation for it. Are you concerned about users would set their analysis threshold below the trigger threshold (for detectors that use triggers)?

JelleAalbers:
No activity for 1.5 years -> closing. Feel free to resurrect this Robert.

PR #269, opened 2022-10-24
Non-asymptotic inference

robertsjames:
Summary [here](https://docs.google.com/presentation/d/15M6HhfCA_TjzDNdPBFNnEav6DXmMh5VtJDu4uPAU2bc/edit#slide=id.p)

This PR adds the opportunity to do inference with Flamedisx when we are in the non-asymptotic regime of the test statistic distributions. It is particularly tailored to the case where we have no shape-varying nuisance parameters


robertsjames:
@makayla-trask @josh0-jrg thanks for agreeing to review. Things have changed a fair bit since the slides I posted at the top of this page outlining the PR, but I think the [LZ template inference tutorial](https://docs.google.com/presentation/d/1LQORLdVJ6eDSz0FszFW4FCpeaMnG1OGiTKHCaKO3RPw/edit#slide=id.p) is a good guide now to the code contained in `non_asymptotic_inference.py`. As for the other changes in this PR, a lot of it is still relevant to inference with the full Flamedisx/FlameNEST sources, and thus future PRs, so I'm going to keep it in. All that stuff is documented from slides 34 onwards in the original slides at the top of this thread

josh0-jrg (commented):
Static module version: Are we sure updates won't be important?

josh0-jrg (approved):

flamedisx/frozen_reservoir.py
josh0-jrg, line 1: Do not give description of the input parameters- but clear given slides and other changes.

flamedisx/likelihood.py
josh0-jrg, line 1: Why not assign constraint_extra_args  in constructor?

flamedisx/likelihood.py
```diff
@@ -398,6 +409,7 @@ def log_likelihood(self, second_order=False,
                     omit_grads=omit_grads,
                     second_order=second_order,
                     empty_batch=empty_batch,
+                    constraint_extra_args=self.constraint_extra_args,
```
josh0-jrg, line 412: Does this work, can it access self her?

flamedisx/likelihood.py
josh0-jrg, line 1: See comment line 412

josh0-jrg (commented):
All looks good but line 412 of flamedisx/likelihood.py has a member variable as the default argument, but self will not exist? Does this work?

flamedisx/likelihood.py
robertsjames: Because you want to set this differently for each toy without re-instantiating the likelihood

flamedisx/likelihood.py
robertsjames: Yep

flamedisx/frozen_reservoir.py
robertsjames: Good spot, thanks - I documented the undocumented inputs

PR #270, opened 2022-10-28
Fix max_dim_sizes bug

robertsjames:
This implements a fix to a bug that's only just come to light. `max_dim_sizes` was, before, a class attribute in `Source`. Blocks in a `BlockSource` could define a `max_dim_size` for their domains, which would then update the `max_dim_sizes` dictionary in the `Source`. The problem comes when you have 2 different types of `BlockSource` that contain blocks which set the `max_dim_size` for the same domain differently. Because `max_dim_sizes` is a class attribute of `Source`, which both `BlockSource`s inherit from, the second source will change the `max_dim_sizes` dictionary for the first source!

This fix makes it an instance attribute instead.

### DIFF
--- a/flamedisx/block_source.py
+++ b/flamedisx/block_source.py
@@ -266,6 +266,7 @@ class BlockModelSource(fd.Source):
         # somewhere else by accident; check none of them appear in multiple blocks
         updated_max_dim_size = []
 
+        self.max_dim_sizes = dict()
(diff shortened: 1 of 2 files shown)


JelleAalbers (approved):
Whoops, mutable defaults strike again. Thanks for spotting Robert!

PR #272, opened 2022-11-10
Rj max sigma update

robertsjames:
This is a small PR at the request of @plt109. It allows blocks to override the default `max_sigma` used in the bounds computation. This is especially important for Pueh's reconstruction bias block, which will use the non-integer stepping functionality. Ideally, the default sources would just move over to the Bayes bounds, negating the need for this slightly. But that's a bit more work, so hopefully this suffices as an interim solution!

### DIFF
--- a/flamedisx/block_source.py
+++ b/flamedisx/block_source.py
@@ -62,6 +62,9 @@ class Block:
     #: for variable tensor stepping
     max_dim_size: ty.Dict[str, int] = dict()
 
+    #: Control the width of the bounds for a particular block
+    max_sigma: ty.Dict[str, float] = dict()
+
     def __init__(self, source):
         self.source = source
         assert len(self.dimensions) in (1, 2), \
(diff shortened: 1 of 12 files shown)


plt109:
Thanks @robertsjames, but do you have an example of how to use this?

robertsjames:
@plt109 you would just put

`max_sigma = {'this_dimension': blah}`

in the relevant block's class attributes, just as you would do for `max_dim_size`

eangelino:
> @plt109 you would just put
> 
> `max_sigma = {'this_dimension': blah}`
> 
> in the relevant block's class attributes, just as you would do for `max_dim_size`

Hi @robertsjames, I am doing some test with this branch. In particular, I would like to reduce the memory usage ,which becomes important for sources at high energy, using this new feature that you propose (specify max_sigma value). But the problem, from what I understood, is that `max_dim_size` keeps the same and need to be adjusted too. Is it correct? 
Wouldn't it be possible to specify these options for different sources?

JelleAalbers:
I think we could close this:
  * The branch is  old enough there are significant merge conflicts.
  * It is a temporary solution, eventually we'd want the Bayes bounds and default sources to work together
  * #273 already allows changing the effective max_sigma in  in the recon bias block by adding multipliers to the `reconstruction_bias_s1_annotate` and `reconstruction_bias_s2_annotate` model functions.  (If you don't do this, it hardcodes the effective max_sigma for the recon bias block at 2.)

Let me know if you disagree @robertsjames. Thanks for making the PR!

PR #273, opened 2022-11-11
Reconstruction Bias Block 2022

plt109:
**What is this reconstruction bias and where does it come from?**
A bias to the size of the S1, S2 signals is introduced during the signal reconstruction process. This bias can come from a variety of sources:
- under-amplified photons [captured as part of the in the single photoelectron(SPE) pulse shape] 
- ADC trigger
- different SPE shape for different PMTs [1 photon will give different areas for different PMTs]
- PMT afterpulses [as PMTs age, they will emit more afterpulses which are merged into the original peak by the reconstruction software]
- by the reconstruction software itself because of the way signals are quantized, peaks are clustered, etc.

Please read [xenon internal document](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:j_angevaare:wfsim_bias) on reconstruction bias for more details.

**How is this reconstruction bias quantified?**
This reconstruction bias, obtained via separate Monte Carlo studies by other analysts(@JoranAngevaare, @FaroutYLq), is defined as
$$\text{reconstruction bias} = \frac{\text{reconstructed area}}{x} - 1$$
where $x$ can either be the raw area hitting the detector, or the reconstructed area from the reconstruction software. 

**How is it implemented in Flamedisx?**
Affects both `ERSource` (see [these lines](https://github.com/FlamTeam/flamedisx/blob/c1dfa9bb4d504ac566a472a68a1c43f049332fa4/flamedisx/lxe_sources.py#L8-L19)) and `NRSource` (see [these lines](https://github.com/FlamTeam/flamedisx/blob/c1dfa9bb4d504ac566a472a68a1c43f049332fa4/flamedisx/lxe_sources.py#L42-L53)).

The reconstruction bias block is implemented as a Gaussian smearing in `ERSource` with zero mean and small standard deviation after the `MakeS1` and `MakeS2` blocks.

Gaussian smearing in `SR1ERSource` with mean and standard deviation obtained from Monte Carlo studies. `_simulate` and `_compute` both use the reconstruction bias and smear maps with $x=\text{raw area}$. `_annotate` uses the reconstruction bias and smear maps with $x=\text{reconstructed area}$.

The reconstruction bias is defined while normalised to 1 (see equation above) in the MC study, but this 1 is added back in [this line](https://github.com/FlamTeam/flamedisx/blob/c1dfa9bb4d504ac566a472a68a1c43f049332fa4/flamedisx/xenon/x1t_sr1.py#L95) for ease of computation since
$$\text{reconstructed area} = (\text{reconstruction bias}+1)\times\text{raw area} $$

**Testing**
Agreement between `_compute` and `_simulate` was tested using an updated [template accuracy notebook](https://github.com/FlamTeam/flamedisx-notebooks/blob/b8acfdf8fcc2c191fe1882706466c442e46a92ed/template_accuracy.ipynb) I got from @robertsjames.
Screenshot from 2022-11-16 11-54-48

A unit test is also added (see [here](https://github.com/FlamTeam/flamedisx/blob/c1dfa9bb4d504ac566a472a68a1c43f049332fa4/tests/test_source.py#L196)).

**Dependencies**
This PR requires the reconstruction bias block to 'sum' over non-integer internal dimensions (`s1_raw`, `s2_raw`) and this functionality is provided in PR #258 by @robertsjames. However, for numerical reasons (see comments in PR #258), it is imperative to be able to tune `max_sigma` for the blocks stepping on non-integer dimensions. This feature is then provided in PR #272 also by @robertsjames.

**Other remarks**
Since this reconstruction bias block affects all sources, I highly encourage @eangelino and @cecilia-ferrari to check for any breaking changes or degradation to fit performance before this PR is merged in.


flamedisx/block_source.py
```diff
@@ -32,6 +32,9 @@ class Block:
     #: _calculate_dimsizes_special().
     bonus_dimensions: ty.Tuple[ty.Tuple[str, bool]] = tuple()
 
+    #: inner_dimensions which take non-integer values
```
robertsjames, line 35: I think all the changes in this file duplicate the PR #258, so we should merge that in first

flamedisx/lxe_blocks/raw_signals.py
```diff
@@ -80,72 +69,50 @@ def _compute(self,
             loc=mean, scale=std + 1e-10
         ).prob(s_observed)
 
+        ''' Think can also remove this chunk
```
robertsjames, line 72: Yes, I don't think you'd want an acceptance on a non-measured signal

flamedisx/lxe_blocks/reconstruct_signals.py
```diff
@@ -0,0 +1,252 @@
...
+            loc=mu,
+            scale=smear)
+        # Call add_extra_columns now, since s1 and s2 are known and derived
+        # observables from it (cs1, cs2) might be used in the acceptance.
+        # TODO: This is a bit of a kludge
+        self.source.add_extra_columns(d)
+        d['p_accepted'] *= self.gimme_numpy(self.signal_name + '_acceptance')
+
+    def _annotate(self, d):
+        # should really be rading reconstruction bias and smear at the raw
```
robertsjames, line 54: I think we discussed this - did you not say that there's a map you can read from to get the bias/smear as a function of reconstructed value also? Though I'm now a bit confused by this, because the map from raw -> final is probabilistic, not deterministic... Could you clarify?

flamedisx/lxe_blocks/reconstruct_signals.py
```diff
@@ -0,0 +1,252 @@
...
+
+        Not quite a dirac delta, which will make this block an identity matrix
+        but computationally intractible or at least not trivially. Need to smear
+        this dirac delta out by a small loading term of 0.001. A larger s1_raw
+        max_dim_size would need a smaller loading term.
+        """
+        return tf.zeros_like(s1_raw, dtype=fd.float_type())+self.s1_smear_load
+
+    # Getting from s1 -> s1_raw
+    def reconstruction_bias_s1_annotate(self, s1_raw):
```
robertsjames, line 160: Similar to comment to that above... Maybe clarify it in the code, also?

flamedisx/source.py
```diff
@@ -58,6 +58,9 @@ class Source:
     #: for which the domain is always a single interval of integers
     no_step_dimensions: ty.Tuple[str] = tuple()
 
+    #: inner_dimensions which take non-integer values
```
robertsjames, line 61: Same as the other comment about duplication

flamedisx/xenon/x1t_sr1.py
```diff
@@ -254,27 +282,85 @@ def set_defaults(self, *args, **kwargs):
...
-                               s2_reconstruction_bias_pivot=\
-                                   DEFAULT_S2_RECONSTRUCTION_BIAS_PIVOT):
-        return calculate_reconstruction_bias(
-            s2,
-            self.recon_map_s2_tf,
-            self.domain_def_s2,
-            pivot_pt=s2_reconstruction_bias_pivot)
-
+    # Not sure what random truth is for
     def random_truth(self, n_events, fix_truth=None, **params):
```
robertsjames, line 364: It's for simulating the truth (energy, position) of the event. Maybe remove this comment?

robertsjames (changes requested):
Looks really good! Just a few minor comments/clarifications

robertsjames:
@plt109 this looks really great. To summarise our discussions offline:

- This validation is for the default sources, which have no bias and the default (small) 'loading' term for the smearing
- This demonstrates that the block, there but essentially doing nothing, does not break things
- You haven't rebased in the per-block `max_sigma` tuning yet - you just hard-coded it for the validations (which is fine!)
- When you tested it with a real case, you found poor agreement in the validation plots (quite a large spread of residuals). But you found exactly the same thing with this block removed for that source, suggesting that there's either a bug there, or more fine-tuning the needs to happen for that source with the `max_dim_size` / `max_sigma` stuff

Do you agree with my summary above? If so, what are the next steps towards getting this merged?

flamedisx/block_source.py
JelleAalbers: Was merged on Jan 19

flamedisx/lxe_blocks/reconstruct_signals.py
JelleAalbers: This just looks like an unfortunate comment. The `annotate` only has to estimate bounds, it is OK that the actual mapping is probabilistic. 

Nonetheless, the code provides the possibility to use different model functions for the raw->recon mapping (used in simulate and compute) and the recon->raw mapping (used in annotate).

flamedisx/lxe_blocks/raw_signals.py
JelleAalbers: Done

flamedisx/lxe_blocks/reconstruct_signals.py
JelleAalbers: Yes, the code defines two pairs of model functions. If we write docs for this it would be good to emphasize this.

flamedisx/xenon/x1t_sr1.py
JelleAalbers: Done

JelleAalbers:
Thanks for the review Robert! I rebased onto main, which includes your non-integer dimension fix; and I removed some unfortunate comments you pointed out.

Merging this won't break any sources not providing maps since the default is no bias and no smearing. It shouldn't increase the computation time either since the new dimension will have a size of 1 element if there is no smearing.

I agree work remains to show that we get a better fit to data with this than without. But the implementation seems right and involved a significant amount of work. If someone resumes work on this, they would want to start from this implementation, not from scratch.

So @robertsjames @plt109 , if neither of you objects, I would propose to merge this.

JelleAalbers (approved):
Looks good to me, I think we can merge and do further testing in analyses.

PR #288, opened 2023-02-06
Xenon012023

cecilia-ferrari:
Ciao! :hugs: 
I think it would be a nice idea to implement the corrections introduced in this branch starting from the main one (at the time of the note) into the recon_bias branch as well. There are just few fixes that can actually solve lots of issues. :smile: 
I also suggest **not deleting** this branch after the merging since it can still be useful for parallel on-going analysis. 
Just one comment: should we wait @eangelino commit about seg map before merging? :wink:


flamedisx/xenon/x1t_sr1.py
```diff
@@ -21,7 +21,7 @@
 DEFAULT_P_DPE = 0.219
 DEFAULT_EXTRACTION_EFFICIENCY = 0.96
 
-DEFAULT_ELECTRON_LIFETIME = 641e3
+DEFAULT_ELECTRON_LIFETIME = 11600000
```
plt109, line 24: Is it a good idea to have such a big default number for elife? But I guess @eangelino sets this via the config file anyway.
```DEFAULT_ELECTRON_LIFETIME = 11600000```

Comment (cecilia-ferrari) [reply flamedisx/xenon/x1t_sr1.py:24]: this is the default value for nT. I modified it in the file because, even if in the config the nT value, it kept resetting the default to the 1t one.. not really understood why.. :thinking:

File: flamedisx/xenon/x1t_sr1.py [hunk 2]
```diff
@@ -298,14 +309,27 @@ def random_truth(self, n_events, fix_truth=None, **params):
         d['z_observed'] = d['z']
 
         # Adding some smear according to posrec resolution
-        d['x_observed'] = np.random.normal(d['x_observed'].values, scale=0.4) # 4 mm resolution)
-        d['y_observed'] = np.random.normal(d['y_observed'].values, scale=0.4) # 4 mm resolution)
+        d['x_observed'] = np.random.normal(d['x_observed'].values, scale=4) # 0.4 = 4 mm resolution)
```
plt109, line 312: @cecilia-ferrari Could you link to some internal xenon note that you got this number from(that position reconstruction uncertainty ~4cm)? From [Jaron's note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:s1_xyz_rec_with_data_driven), it does seem to be on the order of 4cm, but [Matteo's note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:reconstruction_team:s1_xyz_rec_with_data_driven) seems to suggest smt half the value. It's not a huge difference but would be good to document why we increased the posrec smearing by a factor of 10.

flamedisx/xenon/x1t_sr1.py
```diff
@@ -298,14 +309,27 @@ def random_truth(self, n_events, fix_truth=None, **params):
...
+        else:
+            d['drift_velocity'] = self.default_drift_velocity
+            d['z_observed'] = d['z']
         
         # applying fdc
         delta_r = self.fdc_map(
             np.transpose([d['x_observed'].values,
                           d['y_observed'].values,
-                          d['z_observed'].values,]))
+                          d['z_observed'].values/d['drift_velocity'].values,]))
```
plt109, line 332: Is there an accidental double correction here? I suppose the argument of the FDC map is still (x_observed, y_observed, z_observed), but units of `d['z_observed'].values/d['drift_velocity'].values,]` is time?

cecilia-ferrari: When we had a call with Francesco Toschi, we found out that the fdc map requires as input two spatial coordinates and the drift time of the event. that is why we introduced the correction. Now it is not clear if by using both drift_velocity and fdc map we are introducing a tautology.. I hope not.. 
this is for sure something to check.
I tag here @eangelino

flamedisx/xenon/x1t_sr1.py
cecilia-ferrari: Ah! you found a modification that I introduced to improve the fit with not so great results. I do think that at this point this smearing along with the field maps should undergo a review process in order to fix everything with one commit. Tomorrow me and @eangelino will have a call and discuss also about it. commits will be released soon to define once for all the correct implementation :muscle:

plt109: What time's the call, I might be able to join.

flamedisx/xenon/x1t_sr1.py
eangelino: I confirm that this is what we were told by Francesco, which was very non-intuitive. Further, from what I understood, this should be the true drift time (i.e. considering variable drift velocity).

flamedisx/xenon/x1t_sr1.py
```diff
@@ -214,13 +219,29 @@ class SR1Source:
...
     path_drift_field_distortion_correction = 'nt_maps/XnT_3D_FDC_xyt_MLP_v0.2_B2d75n_C2d75n_G0d3p_A4d9p_T0d9n_PMTs1d3n_FSR0d65p.json'
 
+    # Drift velocity map
+    default_drift_velocity = DEFAULT_DRIFT_VELOCITY
+    variable_drift_velocity= True
+    path_drift_velocity='nt_maps/XnT_drift_velocity_map_r_z.json'
+
+    # SE gain map    
+    variable_se_gain= True
+    path_se_gain_map='nt_maps/XENONnT_se_xy_map_v1_mlp.json'
```
plt109, line 229: @eangelino could you link the internal xenon note documenting how this SE gain map was derived and how it should be used/interpreted?

plt109: Also, how does this affect the final interpretation of g2? The information from `se_gain_relative` goes into the `electron_gain_mean` model function but we'll still compute the overall g2 from just the bestfit value for `single_electron_gain`?

flamedisx/xenon/x1t_sr1.py
```diff
@@ -328,6 +365,13 @@ def add_extra_columns(self, d):
                           d['y_fdc'].values,
                           d['z_fdc'].values]))
 
+        if self.variable_se_gain:
+            d['se_gain_relative']=self.se_gain_map(np.transpose([d['x'].values,
```
plt109, line 369: So the SE gain maps are meant to be read from the MC truth positions?

flamedisx/xenon/x1t_sr1.py
cecilia-ferrari: Ok! modified with commit [e85cf9a](https://github.com/FlamTeam/flamedisx/pull/288/commits/e85cf9abdb5ec0a84075be25d8ae32e33e0b147b) 
I am going to resolve this conversation 
Thanks a lot Pueh! :smiling_face_with_three_hearts:

flamedisx/xenon/x1t_sr1.py
eangelino: I am currently looking for a reference note. The map itself contains the the se_gain absolute values in PE for (x,y). Since we would like to keep the single electron gain as a parameter for the fit, the map is normalized by the average value. So, what is interpolated is the relative variation from the average value. This is added in order to consider dishomogeneities even leaving the se_gain average value free.

flamedisx/xenon/x1t_sr1.py
eangelino: Yes, for events at given (x,y) we get a value that allow to scale the default se_gain depending on the position.

flamedisx/xenon/x1t_sr1.py
eangelino: Reference note for se_gain_map: 
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:jianyang:wfsim_data_se_matching

flamedisx/xenon/x1t_sr1.py
cecilia-ferrari: I think we can resolve this review since we found the wiki note and corrected the implementation in fd. 
Cheers :smile:

plt109: Thanks @eangelino for updating this one line https://github.com/FlamTeam/flamedisx/blob/d47218c73fb6d4dbb3c28d183a841b60fa7cc5d4/flamedisx/xenon/x1t_sr1.py#L369

flamedisx/xenon/x1t_sr1.py
plt109: I'm asking Shixiao and the other MC guys about the recommended value to put as the posrec resolution. But since this is a simple change, we can update the value in a later PR.

plt109:
Thanks @cecilia-ferrari and @eangelino for this update! I'm happy to merge this into the [recon bias block branch](https://github.com/FlamTeam/flamedisx/commit/c1dfa9bb4d504ac566a472a68a1c43f049332fa4).

PR #299, opened 2023-03-27
Fix whitespace for flake8

JelleAalbers:
Our builds are currently failing because we missed a trailing whitespace in https://github.com/FlamTeam/flamedisx/pull/285. This removes it.


plt109 (approved):
Thanks @JelleAalbers for maintaining the software standards of fd!

PR #302, opened 2023-04-12
Spatially Varying LZ Source

josh0-jrg:
Spatial varying sources (Xe127, FlatER, DetNR) handled in lz.py
Added:
lz/lz.py:
pickly and histdd
L35
create_spatial_hist() with nested function for processing input [legacy, could change but may be useful in future]
added classes: LZSpatialXe127Source, LZSpatialFlatERSource, and  LZDetNRSource
nest/background_sources.py
added classes: SpatialERSource  and SpatialNRSource


JelleAalbers:
Is this still current? I see there are some changes to the skew Gaussian functions, which might affect other people's work. Would be good to know what these are before merging. (And to fix the tests of course)

JelleAalbers:
Closing since no response -- feel free to resurrect this if relevant!

PR #317, opened 2023-07-31
Template interpolation 

JelleAalbers:
This adds to the intra-bin interpolation support for template sources that @robertsjames added in #269:
  * Support template dimensions other than 2. The `interp_2d` argument is renamed to `interpolate`.
  * Avoid the deprecated interp2d function.
  * Work correctly if the user specifies an events-per-bin template (instead of the default differential rate template).
  * Use linear interpolation instead of a higher-order bivariate spline (which interp2d defaulted to).
  * Add a unit test.

Happy to add an argument to change the interpolation method as well, though I'm not sure what higher-order interpolation would do on noisy histogram templates.


JelleAalbers:
No objections -> merging

PR #318, opened 2023-08-01
Template morphing

JelleAalbers:
This extends flamedisx's support for "old-fashioned" template sources, by adding support for linear / vertical template morphing. The new `MultiTemplateSource` takes multiple (param dict, histogram) combinations. For example:

```python
source = fd.MultiTemplateSource(
    (dict(elife=800, drift_field=120), histogram_1),
    (dict(elife=920, drift_field=150), histogram_2),
    ...)
```
would produce a source that has `elife` and `drift_field` as free parameters, which control a linear interpolation between the differential rate histograms corresponding to those parameters.

Using a rectilinear grid of parameter points is recommended. Something like a cross of points should also work, but internally we interpolate onto a grid anyway on initialization (because tensorflow(_probability) only has rectilinear interpolation)

`MultiTemplateSource` can interpolate  `mu` (expected events) in the same way as the differential rates. Unfortunately, since flamedisx sources usually can't estimate their own mu's, likelihoods still use the mu estimation API (#222). This means you would have to manually specify to use `SimulateEachCallMu` (or `GridInterpolatedMu` (#242) with the same grid as the template morphing source) to get consistent results.

There is a unit test that shows the interpolation appears to be working, but I'd like to test the behavior on likelihoods before we merge.

This PR adds on top of #317, which is much smaller and probably easier to review first. Then I can rebase this.

### DIFF
--- a/flamedisx/source.py
+++ b/flamedisx/source.py
@@ -46,8 +46,8 @@ class Source:
     #: Columns that we don't want to include in the tensor of data columns
     exclude_data_tensor: ty.Tuple[str] = tuple()
 
-    #: Names of array-valued data columns
-    array_columns: ty.Tuple[str] = tuple()
+    #: Array-valued data columns: tuple of (name, length)
+    array_columns: ty.Tuple[ty.Tuple[str, int]] = tuple()
 
     #: Any additional source attributes that should be configurable.
     model_attributes = tuple()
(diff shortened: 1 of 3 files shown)


JelleAalbers:
Robert's commit fixes the (optional, off by default) interpolation inside templates for the bins at the edge of the histogram. We interpolated linearly between bin centers, but I forgot to turn on extrapolation. So you would expect bad stuff to happen if you fed in an event that was closer to the edge of a histogram than the bin center.

Other than that, I saw no objections since last August, so I'm going to merge.

PR #322, opened 2023-08-19
extend `refresh_nt` to grab applefiles as well

plt109:
## What?
This PR extends the function `refresh_nt` to allow it to grab appletree files as well.

## Why?
Some of the maps used by appletree are also used by Flamedisx. This save us the trouble and diskspace of copying the apple tree files over to https://github.com/XENONnT/Flamedisx manually.

We can upload the maps that only Flamedisx uses to our usual place in https://github.com/XENONnT/Flamedisx without going through applefiles.

## How?
A keyword argument `aptf` is added to the function `refresh_nt` [here](https://github.com/FlamTeam/flamedisx/blob/e0c3e499b4ecc506ad4a94e8c6c6f658d65f1c9c/flamedisx/xenon/data.py#L77). The appletree files are cloned if this keyword argument is set to `True`. The default value of this keyword argument is set to `False`.

## Testing?
Tested on Fysikum GPU cluster. But will be nice if @eangelino @cecilia-ferrari can test too.


JelleAalbers:
I get significant conflicts here, probably due to rebasing #273. I suggest we deal with #273 on its own, and make a new commit with this appletree-reading code if needed.

PR #327, opened 2023-09-23
Refactor bounds/dimension stuff into IntegratingSource

JelleAalbers:
**What**: Move all code related to integration and summation (min/max computations, domain stepping, the various `_dimensions` attributions) into a new class named `IntegratingSource`.

**Why**:
  * Readability, this collects related code from many places into one place. 
  * Simplifies our base Source class; `ColumnSource`s, template sources, etc. do not need all of this mechanics.
  * If we want alternatives to our `BlockModelSource` (maybe auto-generating models from equations..) we can now inherit from `IntegratingSource` and keep just the parts we need.

**Related changes**
  * Removed the concept of "additional_bounds_dimensions"; all that did was include `{dimension}_min` and `{dimension}_max` in the columns to keep on the tensorflow side. This was also only used in one place, in the NEST base source for energy. We now just add `energy_min` and `energy_max` to the extra_needed_columns there.
  * Do not populate the `Source.max_dim_sizes` dictionary with defaults. It is accessed in one place, so we can use a dict.get with a default there instead. 
  * The `cap_dimsizes` method was just a few lines and it was only used in one place, so I inlined it. Also refactored the dimsize/stepping calculation a bit.
  * Removed unused argument `_skip_bounds_computation`  for `BlockModelSource._annotate`. `Source.annotate` eats this argument and does not pass it on to `_annotate`.
  * Minor stuff like adding comments or replacing "x = [] then for y in bla append f(y) to x" with `x = [f(y) for y in bla]`

### DIFF
--- a/flamedisx/block_source.py
+++ b/flamedisx/block_source.py
@@ -230,7 +230,7 @@ class FirstBlock(Block):
 
 
 @export
-class BlockModelSource(fd.Source):
+class BlockModelSource(fd.IntegratingSource):
     """Source whose model is split over different Blocks
     """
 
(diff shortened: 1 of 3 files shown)


JelleAalbers:
No objections -> merging.

PR #345, opened 2024-02-02
Discovery code

robertsjames:


JelleAalbers:
This adds capability to do discoveries to Robert's non-asymptotic inference code, see #269. It does not affect other code, and as there are no objections, I will merge this.

PR #360, opened 2024-06-26
Wimprates integration

lorenzomag:
This PR proposes the integration of the package [wimprates](https://github.com/JelleAalbers/wimprates) to automatically compute the energy spectrum of WIMP interactions within FlameNEST for a given set of parameters, including the option to simulate or not signal modulation due to the Earth's revolution around Sun.

Currently, modulation calculation assumes that the timespan in consideration is exaclty 1 year, regardless of the number of bins in time the user chooses.

This was implemented in the class static method `flamedisx.nest.nestWIMPSource.get_energy_hist` which is called during the `__init__()` method of the class itself.

The function `flamedisx.relative_modulation` has been added to `utils.py` and is called during the `get_energy_hist` method.


lorenzomag:
Moved to another PR for another branch

PR #368, opened 2024-08-08
Add wimprates rate generator to nestWIMPSource

lorenzomag:
This PR proposes to utilise the `wimprates` package to compute the differential rates of WIMP-nucleus interaction instead of fetching them from pickle files.

If `modulation==True`, the differential rate (with respect to energy and time, per unit mass) will take into consideration annual modulation patterns, calculated within wimprates itself.

Initial and final timestamps are provided by the user using keyword arguments to the class initialiser method.
The livetime is computed from their difference.

If `modulation==False`, a conservative timestamp is chosen by wimprates to obtain the average annual rate. The differential rate will then be calculated once, and the result repeated into an array.
This option is more time efficient.
A multiprocessing implementation of the former case should be implemented.

**Currently, there is a discrepancy in the output rate by the function wrt the pickled file's energy spectrum by a factor of 10 approximately!**
This needs to be investigated


flamedisx/nest/lxe_sources.py
```diff
@@ -552,23 +555,262 @@ class nestWIMPSource(nestNRSource):
...
+
+        time_bin_edges = (
+            pd.date_range(min_time, max_time, periods=n_time_bins + 1).to_julian_date()
+            - 2451545.0        # Convert to J2000
+        )
+        time_bin_width = wr.j2000_to_datetime(time_bin_edges[1]) - wr.j2000_to_datetime(time_bin_edges[0])
+        time_bin_width = time_bin_width.value * nu.ns
+        times = (time_bin_edges[:-1] + time_bin_edges[1:]) / 2
+
+        scale = time_bin_width / nu.year  # Convert from [per year] to [per time_bin_width]
```
robertsjames, line 655: @lorenzomag I think we need to get rid of this. I tried to explain on Slack (maybe you missed it): whilst in principle keeping this and removing  `/ self.n_time_bins` would solve things at the level of `source.mu_before_efficiencies()`, it will mess up the tensor-based calculation of the differential rate, as the thing that is queried should be the spectrum corresponding to a given event time (such that the sum of the spectrum gives the expected counts within that energy range, after scaling by the exposure). We don't want to break compatibility with the tensor-based calculation

flamedisx/nest/lxe_sources.py
```diff
@@ -552,23 +555,262 @@ class nestWIMPSource(nestNRSource):
...
+        self.n_time_bins = len(self.energy_hist.bin_centers()[0])
+        e_centers = self.energy_hist.bin_centers()[1]
+        self.energies = fd.np_to_tf(e_centers)
+
+        self.array_columns = (("energy_spectrum", len(e_centers)),)
+
+        super().__init__(*args, **kwargs)
+
+    @staticmethod
+    def get_energy_hist(
```
robertsjames, line 747: @lorenzomag can we avoid duplicating all this code?

flamedisx/nest/lxe_sources.py
```diff
@@ -552,23 +555,262 @@ class nestWIMPSource(nestNRSource):
...
+
+        time_bin_edges = (
+            pd.date_range(min_time, max_time, periods=n_time_bins + 1).to_julian_date()
+            - 2451545.0        # Convert to J2000
+        )
+        time_bin_width = wr.j2000_to_datetime(time_bin_edges[1]) - wr.j2000_to_datetime(time_bin_edges[0])
+        time_bin_width = time_bin_width.value * nu.ns
+        times = (time_bin_edges[:-1] + time_bin_edges[1:]) / 2
+
+        scale = time_bin_width / nu.year  # Convert from [per year] to [per time_bin_width]
```
robertsjames, line 777: @lorenzomag same comment as before (though ideally we'll remove the duplicated code)

flamedisx/nest/lxe_blocks/energy_spectrum.py
```diff
@@ -449,7 +449,7 @@ def clip_j2000_times(self, ts):
         return np.clip(ts, tbins[0], tbins[-1])
 
     def mu_before_efficiencies(self, **params):
-        return self.energy_hist.n / self.n_time_bins
+        return self.energy_hist.n
```
robertsjames, line 452: @lorenzomag see my other comment: we want to put this back in

robertsjames (changes requested):

flamedisx/nest/lxe_sources.py
lorenzomag: Hello, yes that is definitely needed. I just wasn't sure where to put the function. Shall I put it in some utils.py file? Or just keep it in there but out of the classes?

lorenzomag: Ok, I refactored the code and made a new module. The name of the module is just the name of the function for now, but we can change it to utils.py if needed, I guess

